PR #1117, opened 2018-03-28
SDK Travis Integration

ehlertjd:
The purpose of the PR is to review the CI changes around SDK. I'm using containers to build swagger documentation (for all builds) and the SDK. SDK Integration tests are running and reporting coverage. I currently disabled pylint for the SDK because it took a long time to run and didn't seem especially useful. Finally, tagged builds deploy SDK artifacts to Google Cloud Storage.

Notable changes to `.travis.yml`:
* Moved build steps into `bin/travis-build.sh` as they were getting more complicated and had conditionals.
* Added cache folders for NPM and Gradle
* Added deployment to GCS for SDK artifacts (will need updated credentials to go live)

It should be safe to ignore `*.mustache` files for the purpose of reviewing this PR. If people are interested in reviewing the full SDK branch before it lands, I'll open a separate PR.


nagem:
To summarize an offline discussion, I am alright with problems in SDK builds and tests marking the core build as failed. The two are separated out visibly in Travis, although not in Github. I would hope in the future we could investigate using a tool like CircleCI that offers this greater level of detail from a Github overview perspective.

nagem:
To clarify, I would rarely (if ever) condone merging a PR that knowingly breaks the SDK, but I wouldn't want to limit

sdk/scripts/docker-tests.sh
```diff
@@ -0,0 +1,152 @@
...
+        case "$1" in
+            -h|--help)
+                log "$USAGE"
+                exit 0
+                ;;
+            -B|--no-build)
+                DOCKER_IMAGE="core:testing"
+				BUILD_SDK_IMAGE=false
+                ;;
+			--python2)
```
ryansanford: Indentation issue

sdk/Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM python:2.7 as python2
```
ryansanford: @ehlertjd What is the value of building these docker images that don't get re-used or persisted? Why not just skip this,  and put the dependency install in `sdk/scripts/tests.sh` then invoke the relevant python images in `sdk/scripts/docker-tests.sh` rather than the one built from this file?

sdk/scripts/docker-tests.sh
```diff
@@ -0,0 +1,152 @@
...
+
+    # Docker build
+    if [ -z "${DOCKER_IMAGE}" ]; then
+        log "Building core:testing ..."
+        docker build -t core:testing ..
+    else
+        docker tag "$DOCKER_IMAGE" "core:testing"
+    fi
+
+	if [ "${BUILD_SDK_IMAGE}" = "true" ]; then
```
ryansanford: Indentation

sdk/scripts/docker-tests.sh
```diff
@@ -0,0 +1,152 @@
...
+    [ $RUN_SHELL ] && SDK_TEST_CMD=bash || \
+                      SDK_TEST_CMD="scripts/tests.sh -- $PYTEST_ARGS"
+    docker run -it \
+        --name ${TEST_PREFIX}-runner \
+        --network ${TEST_PREFIX} \
+		--volume $(pwd):/var/scitran/code/sdk \
+        --env SCITRAN_SITE_API_URL=http://${TEST_PREFIX}-service:9000/api \
+        --env SCITRAN_PERSISTENT_DB_URI=mongodb://${TEST_PREFIX}-service:27017/scitran \
+        --env SCITRAN_CORE_DRONE_SECRET=secret \
+		-w /var/scitran/code/sdk \
```
ryansanford: indentation

sdk/scripts/docker-tests.sh
```diff
@@ -0,0 +1,152 @@
...
+
+"
+
+TEST_PREFIX=sdk-test
+
+main() {
+    local DOCKER_IMAGE=
+    local PYTEST_ARGS=
+    local RUN_SHELL=
+	local DOCKER_TARGET=python3
```
ryansanford: Indentation

sdk/Dockerfile
ehlertjd: The intent is for these images to be included in the docker/image.tar to make future builds faster.

ehlertjd:
@ryansanford Whitespace comments addressed and `~/.vimrc` updated...

Also testing a tagged build for the github releases page.

ehlertjd:
Added migration guide and documentation about the use of Semantic Versioning for this repo.
Ready to merge pending final approval.

sdk/Dockerfile
ryansanford: Thanks. I suspect on the whole there isn't much, if any overall gain, due to the extra 270 MB passed around in the single monolithic docker/image.tar.

This concern is not worth holding up this PR but request follow-up confirmation that the additional complexity  is providing the expected benefit. Maybe that was already done?

PR #1118, opened 2018-03-28
Attempt a semantic conversion of classifications

hkethi002:
### Review Checklist


ehlertjd:
This is based on the table in #402. @lmperry could you review that table for completeness?

bin/database.py
```diff
@@ -1428,14 +1447,23 @@ def upgrade_files_to_45(cont, cont_name):
         modality_container = None
 
         if modality:
-            modality_container = config.db.modalities.find_one({'_id': modality})
+            modality_container = config.db.modalities.find_one({'_id': modality.upper()})
```
nagem: We should do a case insensitive search rather than expecting the modality to be stored in uppercase (even though it will for MR, our only current modality). 

[hierarchy.py](https://github.com/flywheel-io/core/blob/master/api/dao/hierarchy.py#L306) has an example where we do a case insensitive lookup but Mongo might have made changes since that allows for a cleaner way to accomplish the same thing.

bin/database.py
```diff
@@ -1420,6 +1420,25 @@ def upgrade_files_to_45(cont, cont_name):
...
+        "anatomy_inplane":  {"Contrast":"T1", "Intent": "Structural", "Features": "In-Plane"},
+        "anatomy_ir":       {"Contrast":"T1", "Intent": "Structural"},
+        "anatomy_pd":       {"Contrast":"PD", "Intent": "Structural"},
+        "anatomy_t1w":      {"Contrast":"T1", "Intent": "Structural"},
+        "anatomy_t2w":      {"Contrast":"T2", "Intent": "Structural"},
+        "calibration":      {"Intent": "Calibration"},
+        "coil_survey":      {"Contrast": "B1", "Intent": "Calibration"},
+        "diffusion":        {"Contrast":"Diffusion", "Intent":"Structural"},
+        "diffusion_map":    {"Contrast": ["Diffusion", "Fieldmap"], "Intent": "Structural"},
+        "fieldmap":         {"Contrast": "B0", "Intent": "Fieldmap"},
```
ehlertjd: `fieldmap` should be `field_map`

bin/database.py
```diff
@@ -1428,14 +1447,26 @@ def upgrade_files_to_45(cont, cont_name):
         modality_container = None
 
         if modality:
-            modality_container = config.db.modalities.find_one({'_id': modality})
+            modality_container = config.db.modalities.find_one({'_id': re.compile('^{}'.format(re.escape(modality)), re.IGNORECASE)})
```
ehlertjd: @nagem Should we normalize modality to upper case (i.e. `mr` -> `MR`)?

nagem:
Changes cherry-picked into #1010, thanks @hkethi002!

PR #1119, opened 2018-04-02
File modified is updated instead of parent container's

hkethi002:
When a file's info is modified, the file's modified timestamp should be updated.
- Should both be updated?
### Review Checklist


nagem:
LGTM, thanks for finding and addressing this.

PR #1120, opened 2018-04-03
Gear context inputs

ehlertjd:
This is a first-pass, backward compatible implementation of Context Inputs added to flywheel-io/gears#30.

There are currently no required frontend changes since context inputs are populated automatically at job creation time. This PR makes a slight change to how batch proposals are constructed, requiring destination to be resolved at proposal time in order to populate the context input values. 

I have not implemented the endpoint described in flywheel-io/frontend#1442 in this PR.

Tests have been added to cover individual job creation, batch job creation, and rule-based job creation, and I have tested those scenarios via UI.

### Review Checklist

### DIFF
--- a/api/jobs/batch.py
+++ b/api/jobs/batch.py
@@ -6,7 +6,7 @@ import copy
 import datetime
 
 from .. import config
-from ..dao.containerstorage import AcquisitionStorage, AnalysisStorage
+from ..dao.containerstorage import AnalysisStorage
 from .jobs import Job
 from .queue import Queue
 from ..web.errors import APINotFoundException, APIStorageException
--- a/api/jobs/batch.py
+++ b/api/jobs/batch.py
@@ -93,11 +93,11 @@ def find_matching_conts(gear, containers, container_type, context_inputs=False, 
             else:
                 # Create input map of file refs
                 inputs = {}
-                for input_name, files in suggestions.iteritems():
-                    if files[0]['base'] == 'file':
-                        inputs[input_name] = {'type': container_type, 'id': str(c['_id']), 'name': files[0]['name']}
+                for input_name, suggested_inputs in suggestions.iteritems():
+                    if suggested_inputs[0]['base'] == 'file':
+                        inputs[input_name] = {'type': container_type, 'id': str(c['_id']), 'name': suggested_inputs[0]['name']}
                     else:
-                        inputs[input_name] = files[0]
+                        inputs[input_name] = suggested_inputs[0]
                 c['inputs'] = inputs
                 matched_conts.append(c)
         else:
--- a/api/jobs/batch.py
+++ b/api/jobs/batch.py
@@ -6,7 +6,7 @@ import copy
 import datetime
 
 from .. import config
-from ..dao.containerstorage import AnalysisStorage
+from ..dao.containerstorage import AcquisitionStorage, AnalysisStorage
 from .jobs import Job
 from .queue import Queue
 from ..web.errors import APINotFoundException, APIStorageException
@@ -159,6 +159,7 @@ def run(batch_job):
             time_now = datetime.datetime.utcnow()
             analysis_base['label'] = {'label': '{} {}'.format(gear_name, time_now)}
         an_storage = AnalysisStorage()
+        acq_storage = AcquisitionStorage()
 
     jobs = []
     job_ids = []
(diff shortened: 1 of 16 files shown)


ehlertjd:
@nagem My only outstanding question is whether or not we need a DB upgrade for batch objects in the database (since the 'proposal' format changed)

nagem:
@hkethi002 If this goes in before #1082 you'll have a bit of a tough rebase but having the proposal format be a list of jobs will likely make your code simpler.

api/jobs/batch.py
```diff
@@ -165,47 +172,28 @@ def run(batch_job):
...
+            # NOTE: Batch destinations *MUST* be a session or acquisition
+            if job_map['destination']['type'] == 'acquisition':
+                acquisition_id = job_map['destination']['id']
+                session_id = acq_storage.get_container(acquisition_id, projection={'session': 1}).get('session')
+            else:
+                session_id = bson.ObjectId(job_map['destination']['id'])
+
             analysis['job'] = job_map
-            result = an_storage.create_el(analysis, 'sessions', bson.ObjectId(dest['id']), origin, None)
+            result = an_storage.create_el(analysis, 'sessions', session_id, origin, None)
```
nagem: This code will reset the job's destination to the analysis it creates, correct?

api/jobs/batch.py
```diff
@@ -89,7 +94,10 @@ def find_matching_conts(gear, containers, container_type):
                 # Create input map of file refs
                 inputs = {}
                 for input_name, files in suggestions.iteritems():
```
nagem: Maybe we should chose a different variable name for the value on this k,v iterator. It was confusing without going to `suggest_for_files` and seeing the return format of that fuction. `input_name, suggestions`? Something better?

nagem:
Just a small question and suggestion, otherwise looks good. It's too bad we have to have so many checks for file based input (some for maps, some for the FileReference object), but I don't see a better way around it. I appreciate that you looked for file based rather than for context, makes adding other simple input types easier in the future.

api/jobs/batch.py
ehlertjd: That's a good point - this code didn't really change from the previous version, but it seems like it should update the destination.

ehlertjd: I've confirmed that the destination is set by this line.

api/jobs/queue.py
```diff
@@ -164,10 +168,14 @@ def enqueue_job(job_map, origin, perm_check_uid=None):
         inputs = {}
         for x in job_map.get('inputs', {}).keys():
             input_map = job_map['inputs'][x]
-            try:
-                inputs[x] = create_filereference_from_dictionary(input_map)
-            except KeyError:
-                raise InputValidationException('Input {} does not have a properly formatted file reference.'.format(x))
+            base = gear['gear'].get('inputs', {}).get(x, {}).get('base')
```
kofalt: I get real twitchy when we do stuff like this. Are there scenarios where these lookups could fail? If so, why? If not, can we make them a normal lookup?

api/jobs/jobs.py
```diff
@@ -143,7 +143,10 @@ def load(cls, e):
 
             for i in d['inputs']:
                 inp = i.pop('input')
-                input_dict[inp] = create_filereference_from_dictionary(i)
+                if 'type' in i and 'name' in i:
```
kofalt: Should this also check that the type key is set to?

api/jobs/gears.py
```diff
@@ -62,8 +62,10 @@ def add_suggest_info_to_files(gear, files):
 
     schemas = {}
     for x in gear['gear']['inputs']:
-        schema = gear_tools.isolate_file_invocation(invocation_schema, x)
-        schemas[x] = Draft4Validator(schema)
+        input_ = gear['gear']['inputs'][x]
+        if input_.get('base') != 'context':
```
kofalt: This should probably be flipped to == file.

api/jobs/handlers.py
```diff
@@ -460,7 +465,8 @@ def _log_read_check(self, _id):
         if not self.superuser_request:
             if job.inputs is not None:
                 for x in job.inputs:
-                    job.inputs[x].check_access(self.uid, 'ro')
+                    if hasattr(job.inputs[x], 'check_access'):
```
kofalt: Does the `in` keyword not work here? I just haven't used `hasattr` much / before.

api/jobs/handlers.py
```diff
@@ -660,6 +669,12 @@ def post(self):
         if not perm_checked_conts:
             self.abort(403, 'User does not have write access to targets.')
 
+        # For superuser requests, don't check permissions when building context
+        if self.superuser_request:
+            context_uid = None
```
kofalt: Was this basically a bug before this change?

ehlertjd: This only applies to the container checks for the context inputs.

api/jobs/handlers.py
ehlertjd: `hasattr` is a member (attribute) accessor where `in` is a container check. 
In other words, `hasattr` is to `in` as `x.check_access` is to `x['check_access']`

api/jobs/jobs.py
ehlertjd: Kofalt says nvm

ehlertjd:
@kofalt I believe all comments should be addressed - I was able to remove the `.get(key, {})` hijinks by moving a validation check higher up in the function.

ehlertjd:
@nagem I added the endpoint that @kofalt created documentation for (described in flywheel-io/frontend#1442).

This is ready for final review.

PR #1128, opened 2018-04-06
Allow bootstrapping first user without auth when system empty

### LINKED ISSUE #1126: Add bootstrapping capabilities
Add bootstrapping to enable initializing an (empty) API without using a shared secret.

* Add new `bootstrapped: false` config singleton
* Add handler for `POST /bootstrap`, which
   * enables creating the first user (with an API key) `if not config['bootstrapped']`
   * sets `config['bootstrapped'] = true`
* Update all bootstrapping code to use the method above (core, fly/fly, sdk, reaper, etc.)

ambrussimon:
Resolves #1126 
* `POST /users` instead of a new endpoint (makes sense to me, but can be convinced about separation)
* Re-running `bootstrap-accounts.sh` requires an API key

Pending review:
* [ ] flywheel-io/flywheel#601 @ryansanford 

Pending work:
* [ ] flywheel-io/installer @ambrus

### DIFF
--- a/api/handlers/userhandler.py
+++ b/api/handlers/userhandler.py
@@ -111,11 +111,26 @@ class UserHandler(base.RequestHandler):
         payload['root'] = payload.get('root', False)
         payload.setdefault('email', payload['_id'])
         payload.setdefault('avatars', {})
+
+        if self.public_request and config.db.users.count() == 0:
+            try:
+                config.db.singletons.insert_one({'_id': 'bootstrap', 'uid': payload['_id']})
+            except pymongo.errors.DuplicateKeyError:
+                pass
+            else:
+                payload['root'] = True
+                result = mongo_validator(self.storage.exec_op)('POST', payload=payload)
+                if result.acknowledged:
+                    api_key = UserApiKey.generate(payload['_id'])
+                    return {'_id': result.inserted_id, 'key': api_key}
+                else:
+                    config.db.singletons.delete_one({'_id': 'bootstrap'})
+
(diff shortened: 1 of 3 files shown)


ryansanford:
Thanks @ambrussimon 
If we cannot specify an API key for this first user, we need a mechanism to get that key out of load_users.py, or maybe the creation of that user should be it's own script? That would allow us to migrate away from using the drone secret for automation of subsequent bootstrapping/data loading activities.

@nagem 
Can you confirm with Ashley that this user being able to be assigned access to groups/projects like any other user is ok, and if it is, the appropriate default first/last/email values?

nagem:
@ryansanford @gsfr and I had a quick chat about this today - we believe the caller should send a payload to create the first user where they specify details about a real intended user of the system. The API will generate an API key for that newly created user (and return it in the response to the `POST` request) and that API key can be used to add additional users and devices. Changes will need to be made to infrastructure to take the API key created and use it in further bootstrapping (https://github.com/flywheel-io/flywheel/issues/592)

AshleyAncona:
Waiting on a flyfly change

ambrussimon:
Updated PR desc for tracking progress.

ambrussimon:
@nagem @ryansanford I limited the change scope of this PR to the addition of the feature only (== not touching `load_users_drone_secret.py`). That is backwards-compatible, tested, and already approved, thus I'm going ahead and merging this once green. What's left is actually transitioning to using it throughout the whole infra, which is likely going to be 3 PRs needing merge coordination in `core`, `fly` and `installer`.

PR #1132, opened 2018-04-09
Added SSEHandler to workaround UWSGI limitations

ehlertjd:
This commit spawns a thread for requests that need to send SSE events.
The thread then uses UWSGI offload threads to respond via Unix Socket.

See discussion [here](http://uwsgi-docs.readthedocs.io/en/latest/articles/OffloadingWebsocketsAndSSE.html) for why this is EVIL<sup>TM</sup>. 

One possible improvement would be to add a supervisor for SSE threads.

### Review Checklist

- Tests were added to cover all code changes
- Documentation was added / updated
- Code and tests follow standards in CONTRIBUTING.md

PR #1133, opened 2018-04-10
Fix SSE by bypassing webapp2 handling

### LINKED ISSUE #946: SSE endpoints are not delivered correctly
<a href="https://github.com/kofalt"></img></a> **Issue by [kofalt](https://github.com/kofalt)**
_Wednesday Oct 04, 2017 at 20:37 GMT_
_Originally opened as https://github.com/scitran/core/issues/946_

----

Dan and I observed a new bug in the SSE endpoint `/projects/x/packfile-end`. All responses are not delivered in real-time as they should be, and instead are buffered and delivered all at the end.

This causes two problems. One, the progress or other live events are not delivered to the user in a timely fashion, and two, the gateway will time out if the operation takes too long because no traffic is being sent.

You can observe this by adding `time.sleep(1)` to `PackfilePlacer.finalize`.

Ryan, my initial thought is this might be an gateway problem. Nginx response buffering?

ehlertjd:
SSE Fix, now with 20% more sanity than #1132 ! 

By returning a callable from our request handler, we can take over sending SSE events with the `start_response` call. Tested with a 6-minute long SSE stream. Also tested that client disconnect does not interrupt handling.

Closes #946.

### Review Checklist

### DIFF
--- a/api/web/start.py
+++ b/api/web/start.py
@@ -82,6 +82,9 @@ def dispatcher(router, request, response):
     try:
         rv = router.default_dispatcher(request, response)
         if rv is not None:
+            # Bypass webapp2 handler, rv will be called with (environ, start_response)
+            if callable(rv):
+                return rv
             response.write(json.dumps(rv, default=encoder.custom_json_serializer))
             response.headers['Content-Type'] = 'application/json; charset=utf-8'
     except webapp2.HTTPException as e:
--- a/api/upload.py
+++ b/api/upload.py
@@ -140,24 +140,30 @@ def process_upload(request, strategy, container_type=None, id_=None, origin=None
     if placer.sse and not response:
         raise Exception("Programmer error: response required")
     elif placer.sse:
-        response.headers['Content-Type'] = 'text/event-stream; charset=utf-8'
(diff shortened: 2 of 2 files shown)


nagem:
Adding @ambrussimon as a reviewer as well.

api/web/start.py
```diff
@@ -82,6 +82,9 @@ def dispatcher(router, request, response):
     try:
         rv = router.default_dispatcher(request, response)
         if rv is not None:
+            # Bypass webapp2 handler, rv will be called with (environ, start_response)
+            if callable(rv):
+                return rv
```
ambrussimon: This is pure gold. Kudos, I was looking at webapp2 sources for a day and didn't find where and how I could get a hold of the raw uwsgi `start_response`.

kofalt:
To clarify, you can confirm that if the TCP connection drops, processing will run to completion?

ehlertjd:
@kofalt correct

api/web/start.py
kofalt: It's definitely one of the least discoverable codebases I have read in the last few years...

PR #1139, opened 2018-04-12
Enable bind-mounting whole repo in dev/tests

### LINKED ISSUE #1138: Enable bind-mounting the whole repo during development
* Now installing core as a package in [Dockerfile](https://github.com/flywheel-io/core/blob/43eedae604c4e8db994b5bcaf3495d81e0200bcc/Dockerfile#L63)
* That creates dir `/var/scitran/code/api/core.egg-info` which is required for the imports to work
* This dir is _not_ present on the host, bind-mounting the whole repo to `/var/scitran/code/api` doesn't work because it shadows the egg-info dir

For development purposes, bind-mounting the whole repo would be way more intuitive and less error-prone. Find and implement a suitable workaround to enable just that.

Ideas so far:
* `docker cp` egg-info to host after every build (+ .gitignore)
* create egg-info elsewhere: https://thekev.in/blog/2016-11-18-python-in-docker/

ambrussimon:
Resolves #1138 

### Review Checklist

### DIFF
--- a/tests/bin/docker-tests.sh
+++ b/tests/bin/docker-tests.sh
@@ -51,7 +51,6 @@ main() {
         shift
     done
 
-    # Docker build
     if [ -z "${DOCKER_IMAGE}" ]; then
         log "Building flywheel/core:testing ..."
         docker build -t flywheel/core:testing .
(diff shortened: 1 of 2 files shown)


ehlertjd:
My only concern with this approach is whether or not we would need to rebuild egg-info if we add or rename files to the project. (i.e. if I create a new source file, do I need to re-run pip install to update 
the sources list, or does it just work?)

Does this need a corresponding `.gitignore` change?

ambrussimon:
New source files work without rebuilding [the egginfo].

I found an egg-info exclusion already present in .gitignore, now it's used
[again?]

tests/bin/docker-tests.sh
```diff
@@ -70,11 +69,14 @@ main() {
         flywheel/core:testing \
         mongod
 
+    rm -rf core.egg-info
+    docker create --name core-test-egg-info flywheel/core:testing
+    docker cp core-test-egg-info:/var/scitran/code/api/core.egg-info ./
```
ryansanford: I generally like the pattern of `docker run --rm -v "$(pwd):/pwd" flywheel/core:testing cp /var/scitran/code/api/core.egg-info /pwd` due to it being a one-liner, and avoids the need for cleanup. Not a critical change.

PR #1145, opened 2018-04-17
Log file replacement

### LINKED ISSUE #1129: File replacement is not tracked in user access log
File replacement is logically equivalent to (modify) or (delete + create). Modify and delete are subject to the user access log, and thus file replace should also be.

nagem:
Closes #1129

We've really pushed the limits of the original intention of access logging (logging user access to data). Now that we are adding more complicated actions, it would be worth it to redesign how user action logging should work in the future. For now we'll pass the callable to non-handler layers.

### Review Checklist

### DIFF
--- a/api/dao/hierarchy.py
+++ b/api/dao/hierarchy.py
@@ -12,7 +12,9 @@ from .. import config
 from .basecontainerstorage import ContainerStorage
 from ..auth import has_access
 from ..web.errors import APIStorageException, APINotFoundException, APIPermissionException
+from ..web.request import AccessType
 from . import containerutil
+
 
 log = config.log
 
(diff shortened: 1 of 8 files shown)


api/dao/hierarchy.py
```diff
@@ -249,6 +251,10 @@ def upsert_fileinfo(cont_name, _id, fileinfo):
         fileinfo['created'] = fileinfo['modified']
         container_after = add_fileinfo(cont_name, _id, fileinfo)
     else:
+        if fileinfo.get('hash') and file_before.get('hash') != fileinfo.get('hash'):
```
nagem: This is looking for hash's existence because sometimes this method is called to update file db object fields only. That is change in #1082 to be separate functions.

nagem:
Attn: @ryansanford

PR #1151, opened 2018-04-19
Add device CRUD handlers and device key auth

### LINKED ISSUE #1127: Device API keys
* Extend/tune devices endpoints
   * site admin can `POST /devices {"type": "reaper (required str)", "name": "foo (opt str)"}`
   * "new-style" device._id is `ObjectId`, and device.api_key is also auto-generated
   * devices can update themselves (eg. set name, add error msgs, etc.) via `PUT /devices/self`
   * `/devices/self` continues to double as heartbeat, too
* Add migration script for old-style devices
* Implement device key authentication
* Update clients to use it
   * Reaper
   * Engine

ambrussimon:
Resolves #1127 

Changes backwards-compatible with shared secret **auth headers**
**Breaking changes** on the device db schema and on `/devices` API endpoints
* rename `._id(str)` to `.label`, `.method` to `.type` (eg. engine, reaper, etc. - not an enum _yet_)
* `._id(ObjectId)` is generated for new and existing devices alike
* Relevant PRs:
   * flywheel-io/reaper#130
   * flywheel-io/flywheel#573 (prometheus dev integration)
   * engine _insert link_

#### New device workflow
* Admins can `POST /devices {"type": str(opt), "label": str(opt)}`
* ApiKey auto-generated and shown _for admins_ `GET /devices` and `/devices/<device_id>`
* Device key usage == user api key usage == `Authorization: scitran-user <device_key>`
* Grants same privileges as old `drone_request` _for backward comp. for now_
* Devices can check in via `PUT /devices/self` (payload and all keys are optional)
```{"type": str, "name": str, "interval": int, "info": object, "errors": list(str)}```

Pending review:
* [ ] flywheel-io/flywheel#573 @ryansanford 
* [x] flywheel-io/reaper#130 @gsfr

### DIFF
--- a/api/api.py
+++ b/api/api.py
@@ -30,11 +30,11 @@ routing_regexes = {
     # Group ID: 2-32 characters of form [0-9a-z.@_-]. Start and ends with alphanum.
     'gid': '[0-9a-z][0-9a-z.@_-]{0,30}[0-9a-z]',
 
-    # Container ID: 24-character hex
-    'cid': '[0-9a-f]{24}',
-
     # User ID: any length, [0-9a-z.@_-]
     'uid': '[0-9a-zA-Z.@_-]*',
+
+    # Object ID: 24-character hex
+    'oid': '[0-9a-f]{24}',
 
     # Container name
     'cname': 'groups|projects|sessions|acquisitions|collections|analyses',
@@ -50,9 +50,6 @@ routing_regexes = {
 
     # Filename/classification
     'fclass': '.+(?=/classification)',
-
-    # Note ID
-    'nid': '[0-9a-f]{24}',
 
     # Schema path
     'schema': r'[^/.]{3,60}/[^/.]{3,60}\.json'
@@ -156,12 +153,12 @@ endpoints = [
         route('/gears',                                  GearsHandler),
         route('/gears/check',                            GearsHandler, h='check', m=['POST']),
         route('/gears/temp',                             GearHandler, h='upload', m=['POST']),
-        route('/gears/temp/<cid:{cid}>',  GearHandler, h='download', m=['GET']),
+        route('/gears/temp/<cid:{oid}>',  GearHandler, h='download', m=['GET']),
         prefix('/gears', [
             route('/<:[^/]+>',                           GearHandler),
             route('/<:[^/]+>/invocation',                GearHandler, h='get_invocation'),
             route('/<:[^/]+>/suggest/<:{cname}|subjects>/<:[^/]+>', GearHandler, h='suggest'),
-            route('/<:[^/]+>/context/<:{cname}>/<:{cid}>', GearHandler, h='get_context', m=['GET']),
+            route('/<:[^/]+>/context/<:{cname}>/<:{oid}>', GearHandler, h='get_context', m=['GET']),
         ]),
 
         # Batch jobs
@@ -181,8 +178,8 @@ endpoints = [
         route( '/devices',              DeviceHandler,                 m=['POST']),
         prefix('/devices', [
             route('/status',            DeviceHandler, h='get_status', m=['GET']),
-            route('/self',              DeviceHandler, h='get_self',   m=['GET']),
-            route('/<device_id:[^/]+>', DeviceHandler,                 m=['GET']),
+            route('/self',              DeviceHandler, h='put_self',   m=['PUT']),
+            route('/<device_id:{oid}>', DeviceHandler,                 m=['GET', 'DELETE']),
         ]),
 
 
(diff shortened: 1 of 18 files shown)

### CODE REVIEW

**Comment by kofalt** [comment]
Details in PR description appreciated :)

File: api/auth/apikeys.py [hunk 1]
```diff
@@ -6,6 +6,7 @@
 
 log = config.log
```
nagem: I liked how you organized this with class variables, nice work.

api/handlers/devicehandler.py
```diff
@@ -22,63 +23,85 @@ class DeviceHandler(base.RequestHandler):
...
         super(DeviceHandler, self).__init__(request, response)
-        self.storage = containerstorage.ContainerStorage('devices', use_object_id=False)
+        self.storage = containerstorage.ContainerStorage('devices', use_object_id=True)
 
     @require_login
     def get(self, device_id):
-        return self.storage.get_container(device_id)
+        device = self.storage.get_container(device_id)
+        if self.user_is_admin:
+            self.join_api_key(device)
```
nagem: @ryansanford Are we alright with all site admins having access to device API keys? Or do we need another layer in our user roles breakdown?

api/api.py
```diff
@@ -173,15 +170,15 @@ def prefix(path, routes):
         route( '/devices',              DeviceHandler,                 m=['POST']),
         prefix('/devices', [
             route('/status',            DeviceHandler, h='get_status', m=['GET']),
-            route('/self',              DeviceHandler, h='get_self',   m=['GET']),
-            route('/<device_id:[^/]+>', DeviceHandler,                 m=['GET']),
+            route('/self',              DeviceHandler, h='put_self',   m=['PUT']),
```
nagem: It looks like we're getting rid of the ability for devices to look themselves up without knowing their ObjectId. Do any devices do this in the field? Is there a need for this functionality? If no for both, then we're fine.

nagem: I think I original wrote it to mimic `users/self`, so I'm not 100% sure if it was every used/wanted.

ambrussimon: A grep including all relevant repos did not yield anything. I'm pretty certain nothing used it ever. This reminded me of updating swagger though, thanks!

nagem:
LGTM. 

@ryansanford I'm assuming the answer to this question is yes:

>Are we alright with all site admins having access to device API keys? Or do we need another layer in our user roles breakdown?

api/api.py
nagem: Eventually the engine will have to make requests with it's API key rather than the shared secret, as we are phasing support of that method of auth out. It will need to know that but won't need to know which device `_id` that is attached to.

api/handlers/devicehandler.py
```diff
@@ -22,63 +23,85 @@ class DeviceHandler(base.RequestHandler):
...
     def get_status(self):
-        devices = self.storage.get_all_el(None, None, None)
-        response = {}
         now = dt.datetime.now()
-        for d in devices:
-            d_obj = {}
-            d_obj['last_seen'] = d.get('last_seen')
+        statuses = {}
+        for device in self.storage.get_all_el(None, None, None):
+            status = {'last_seen': device.get('last_seen')}
```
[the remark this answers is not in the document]
nagem: Now it's right here: https://github.com/flywheel-io/core/pull/1151/files#diff-c86285960fea45f6471319c63cba0431R105

Used to be when setting origin: https://github.com/flywheel-io/core/pull/1151/files#diff-c86285960fea45f6471319c63cba0431L284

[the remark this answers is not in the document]
nagem: Short answer: yes

api/api.py
kofalt: Yeah, a different format key is fine.

ambrussimon: Confirmed.

api/handlers/devicehandler.py
```diff
@@ -22,63 +23,85 @@ class DeviceHandler(base.RequestHandler):
...
+            self.join_api_key(device)
+        return device
 
     @require_login
     def get_all(self):
-        return self.storage.get_all_el(None, None, None)
+        devices = list(self.storage.get_all_el(None, None, None))
+        if self.user_is_admin:
+            for device in devices:
+                self.join_api_key(device)
```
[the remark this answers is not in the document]
ambrussimon: Good point, implemented via `?join_keys=true`.

swagger/examples/output/device-list.json
```diff
@@ -1,45 +1,45 @@
 [
   {
+    "_id": "000000000000000000000000",
```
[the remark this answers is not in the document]
ambrussimon: I'm not sure I follow. A `GET /devices` includes the unique ObjectId `_id` with each entry, and that `_id` is also used in the endpoint `GET /devices/<device_id>`. I would assume API clients find that unique id necessary/useful and is thus reflected in the output examples.

PR #1157, opened 2018-04-23
Fix batch jobs with context inputs not running

ehlertjd:
Found while testing context-inputs on classification-edit branch. 

Batch scheduling a job with context-inputs results in an execution error because each input is treated as a file ref. This is a simple fix, and I verified that the jobs run successfully after this.

### Review Checklist

### DIFF
--- a/api/jobs/jobs.py
+++ b/api/jobs/jobs.py
@@ -362,11 +362,12 @@ class Job(object):
             for input_name in self.inputs.keys():
                 i = self.inputs[input_name]
 
-                r['inputs'].append({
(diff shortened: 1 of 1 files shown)


nagem:
Do you think it's worth it to test the response format from `jobs/next`? Not as part of this PR but maybe as a low hanging fruit for integ testing updates. With config, API keys, files and context to ensure they're formatted how the engine will expect them. I'll open an issue.

ehlertjd:
@nagem Yes, I think it would be good to try to cover that endpoint with integration testing.

PR #1164, opened 2018-04-25
Download labels will replace / with : on download

### LINKED ISSUE #1152: Downloads of containers with `/` in their labels causes unintentional folder levels
Downloading a session named `04/21/16 16:58 PM` using `GET api/download` results in this folder structure:

screen shot 2018-04-19 at 12 54 16 pm

hkethi002:
Fixes #1152 
### Review Checklist


nagem:
>Would it make sense to use sanitize_string_to_filename in util.py to sanitize labels

I agree, I think we use it for analyses labels already (but I might be misremembering).

PR #1171, opened 2018-04-27
Prevent adding permissions for unknown users

### LINKED ISSUE #1143: Adding a user permission via SDK can cause issues
If that user does not exist in the system it will cause the app to throw the following:

```
2018-04-13 16:50:16      scitran.api                  base.py   375:ERRO Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/webapp2.py", line 570, in dispatch
    return method(*args, **kwargs)
  File "./api/handlers/containerhandler.py", line 365, in get_all
    modified_results = self.join_user_info(modified_results)
  File "./api/handlers/containerhandler.py", line 207, in join_user_info
    user = users[uid]
KeyError: u'imperry@stanford.edu'
 request_id=339cbf24-1523638216
```

This causes the UI not to load with a 'server error occurred' message. 

Suggest that the endpoint validate that the user exists and reject the request if not. 

cc: @ehlertjd

hkethi002:
Fixes #1143 

Prevent adding permissions for users that don't exists.

### Review Checklist

### DIFF
--- a/api/handlers/listhandler.py
+++ b/api/handlers/listhandler.py
@@ -13,7 +13,8 @@ from ..auth import listauth, always_ok
 from ..dao import noop
 from ..dao import liststorage
 from ..dao import containerutil
-from ..web.errors import APIStorageException, APIPermissionException
+from ..dao import containerstorage
+from ..web.errors import APIStorageException, APIPermissionException, APIUnknownUserException
 from ..web.request import AccessType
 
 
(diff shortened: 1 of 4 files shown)


api/handlers/listhandler.py
```diff
@@ -208,8 +209,13 @@ class PermissionsListHandler(ListHandler):
     """
     def post(self, cont_name, list_name, **kwargs):
         _id = kwargs.get('cid')
-        result = super(PermissionsListHandler, self).post(cont_name, list_name, **kwargs)
+
         payload = self.request.json_body
+        user_ids = [user['_id'] for user in containerstorage.ContainerStorage('users', use_object_id=False).get_all_el({}, None, None)]
```
ehlertjd: I'd create a utility function `user_exists` and use either:
`containerstorage.ContainerStorage('users', use_object_id=False).get_all_el({id: payload.get('_id')}, None, {'_id':1})]`
or just `config.db.users.find_one({_id: payload.get('_id')}, {'_id':1})`

nagem:
I assume `PUT` will 404 if they user doesn't exist in the list? If so - LGTM!

hkethi002:
Yes, regardless of wether the user exists or not

PR #1173, opened 2018-04-27
Separate concerns in access log and add bulk insert

ehlertjd:
Added a create_entry function to create and validate access log entries.
Also added a bulk_log_access routine to bulk insert access log entries.

### Review Checklist

### DIFF
--- a/api/access_log.py
+++ b/api/access_log.py
@@ -1,0 +1,127 @@
+import datetime
+
+from . import config
+
+from .dao.hierarchy import get_parent_tree
+from .web.request import AccessType
+
+
+def log_user_access(request, access_type, cont_name=None, cont_id=None, filename=None, multifile=False, origin=None, download_ticket=None):
+    """Create a single access log entry, populating context automatically.
+
+    Only creates one access log entry for multiple downloads of a single file via ticket.
+    Context will be created based on passed in container name and id.
+
+    Arguments:
+        request (object): The request object
+        access_type (AccessType): The access log entry type
+        cont_name (str): The name of the container collection being accessed
+        cont_id (str): The id of the container being accessed
+        filename (str): If this is a file access, the name of the file accessed.
+        multifile (bool): Whether or not this is a multi file download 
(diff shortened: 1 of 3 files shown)


api/access_log.py
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    if context is None:
+        if access_type not in [AccessType.user_login, AccessType.user_logout]:
+            raise Exception('Programmer Error: Log context MUST be specified!')
+    else:
+        result['context'] = context
+        if access_type in [AccessType.view_file, AccessType.download_file, AccessType.replace_file, AccessType.delete_file]:
+            if not context.get('file'):
+                raise Exception('Programmer Error: File log entries must include file')
```
nagem: Do we want to do any more validation that the context passed is properly formatted? Or is trusting the caller enough? Or requiring extensive automated testing of the callers? Don't want to find out our logs were missing key context after the fact. This also might be a larger question about gauging the reliability of our access logging.

nagem: Right now we are relying on `get_parent_tree` to provide most of the information and could make improvements on testing it as extensively as we could be, but the context creation is all in one place.

ehlertjd: My intention is that the feature that will make use of the bulk insert and will have all of the context on-hand for access logging, so it would be performance impacting to do the context lookup for each record. 

It's true that every existing request the context is created from the container tree in a reliable way, so I don't have much concern there. 

I could see some value in adding an additional sanity check that each specified context container is a valid container name and has an id and label - I'm happy to make that change for this PR if you would agree that it's valuable. But ultimately, I think we're stuck relying on the caller to build the correct context and include the correct set of containers.

nagem: I think I would be happy with additional testing of the logging, a schema check (json or otherwise) is probably overkill. It's an idea we can come back to if our priorities or confidence change.

ehlertjd: I'll make sure that future features that I add which require access logging have tests covering the entries created.

PR #1174, opened 2018-04-27
Api version

szilveszterb:
### Review Checklist

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -238,7 +238,19 @@ def get_public_config():
     }
 
 def get_version():
-    return db.singletons.find_one({'_id': 'version'})
+    version_object = db.singletons.find_one({'_id': 'version'})
+    if not version_object:
+        return version_object
+
+    try:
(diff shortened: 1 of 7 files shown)


AshleyAncona:
@ehlertjd taking this PR.

api/config.py
```diff
@@ -238,7 +238,19 @@ def get_public_config():
...
+    if not version_object:
+        return version_object
+
+    try:
+        with open('/var/scitran/api_version.txt') as f:
+            api_version = f.read()
+            version_object['release'] = api_version
+    except IOError:
+        version_object['release'] = ''
+
```
nagem: Could we persist this to the version db entry on startup rather than reading it from a file at call time? Two benefits there: 
  - It is easy to match a proper core version to a db backup
  - It is available to other methods within the API 

If that isn't feasible, it would be nice to have it as a function under `util.py` to satisfy the second bullet point.

ehlertjd:
@ryansanford Added to `.gitignore`

PR #1179, opened 2018-05-01
Sdk version check

### LINKED ISSUE #1176: SDK2 Version Check
Unless specified otherwise (through environment variable or config option) SDK2 clients should check major.minor version against the server and print out a warning to the user to proceed at their own risk.

In addition, if either the client or server aren't configured with a semantic version (or the version is missing) print out a separate warning that the user is running in an unsupported way (i.e. with development versions of the software)

Depends on #1174

ehlertjd:
Adds User-Agent string to SDK2 Clients, and check that client/server versions match, issuing a warning if they do not.

Closes #1176.

### Review Checklist


kofalt:
I'm rather allergic to actions being taken in constructors. Maybe a warn on first network request instead?

kofalt:
This also looks like it would warn / nag on every release, and we don't necessarily have anything new for the SDK on every release. What would you think about a new `API Level` integer that is incremented the same way we increment the DB version, and changes more rarely than a release identifier?

ehlertjd:
@kofalt Agree on your first point - I think that's doable, will look into it.

For an `API Level` integer, would you really expect it to increment less frequently than the minor version number? Who is responsible for incrementing it? Does a greater version number mean that there is new functionality, or that there was a breaking change (or both)?

I would rather that we follow semver principles. If we can reliably do that, then I could see an argument for only printing the warning if the major versions differ, or if the SDK minor version is greater than the API minor version (indicating that the SDK may have features not yet supported by server version.)

kofalt:
I'm not familiar enough with when / how often the API's semver changes, so I don't know. I would expect there to be sequences of releases that some customers receive that would not require an SDK change - would those two releases have a bumped minor version?

If the semver changes reliably IFF there's an SDK change required, then I would agree that a major versions differ or similar would be a good way to do it. What I'm trying to avoid is a docker/ruby/whatever-ish scenario where a scientist user (who probably is running the exact same script they wrote X months ago) is constantly bugged about software that is "out of date".

kofalt:
That would be one option, yeah. Is that a terrible idea? I'm trying to feel out how we can make sure we interrupt the user as few times as are necessary.

Another idea: warn (maybe even error) on major version mismatch, and offer & doc a `check_update` func alongside the basic API key instructions. Let the user control their own destiny - I would expect most people to stick to what they know, and the exception cases are the ones were in tight communication with about new feature X.

ehlertjd:
So at this point, I think we're arguing about whether the better approach is to opt-in or opt-out to printing version warnings if the user finds themselves in an unsupported scenario due to a minor version difference. It sounds like you agree that we should always check and warn on a major version difference.

From a use-case perspective, this is my current line of thinking:

Let's say that my server version is 2.1.3 and the latest SDK version is 2.2.0. I'm starting a new project, so I `pip install flywheel-sdk` which gives me the latest (2.2.0) version, which may have some minor features that are not present on my server. While developing, I execute my first request `self = fw.get_current_user()` and see the warning:

```
Client and server versions do not match. This is an unsupported configuration!
Use "pip install flywheel-sdk~=2.1.3" 'to install a compatible version for this server.
```

So I follow the instructions, finish creating my application with no further warnings. From that point, even if the server goes through a number of patches and minor upgrades, the script never warns about a potential version mismatch. 

When a major version upgrade happens, the script either warns or errors (I would argue warn because we can't really predict what's going to break). 

If I really didn't want to downgrade my SDK version, I could add `skip_version_check=True` to the constructor and we wouldn't print minor version differences - I'd argue that we should still check and warn on major version differences.

ryansanford:
My advice:
- We should not further complicate our versioning strategy at this stage. Let's go through the exercise for a while and lean on that experience to make adjustments.
- Mismatch on MAJOR should produce a warning. I don't think anyone is debating this.
- Mismatch on MINOR should produce a message to upgrade SDK to ensure you have access to the full range of capabilities. The gravity of the message should be lighter than a MAJOR mismatch. I'd consider this a "user outreach" feature to foster engagement from users we don't know about. If there is strong consensus that the noise is too much of a negative, I'm all ears. If nothing else, this would be a trivial thing to remove going forward.

PR #1181, opened 2018-05-01
Add util function for creating Content-Range header response

nagem:
When a client would request a byte range with no `last` byte, the `Content-Range` header generated by the API would be malformed. For example:

Client request header:
```
Range: bytes=182190080-
```

API response header:
```
Content-Range: bytes 182190080-None/318839180
```

@davidfarkas please review the changes I made and make any additional/different changes you deem fit. These changes fixed an issue with jumping to mid-video in our video viewer but I did not spend much time comparing my results to the rfc for range reads. I will send you an email will more details on how to recreate the issue locally for testing. 

### Review Checklist

### DIFF
--- a/api/util.py
+++ b/api/util.py
@@ -317,6 +317,17 @@ class RangeHeaderParseError(ValueError):
     """Exception class representing a string parsing error."""
 
 
+def build_content_range_header(first, last, size):
+    if first == None:
+        first = 0
+    if last == None:
+        last = size-1
+
+    print 'first is {} last is {} size is {}'.format(first, last, size)
+
+    return 'bytes %s-%s/%s' % (str(first), str(last), str(size))
+
+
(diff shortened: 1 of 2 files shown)


api/util.py
```diff
@@ -317,6 +317,17 @@ class RangeHeaderParseError(ValueError):
     """Exception class representing a string parsing error."""
 
 
+def build_content_range_header(first, last, size):
+    if first == None:
+        first = 0
+    if last == None:
+        last = size-1
```
davidfarkas: Oh, yep, I forgot to set the last byte when the `parse_range_header` return with `None`. 
Actually, only the last byte can be None, the first byte always has a value, otherwise the  `parse_range_header` raises an exception, so the first `if` is not necessary.

nagem: Thanks, I'll remove it to prevent any confusion.

PR #1182, opened 2018-05-02
Must have site admin to search all data

### LINKED ISSUE #1072: Remove ability to view "all_data" without site admin privs in search
_Wednesday Feb 07, 2018 at 22:11 GMT_
_Originally opened as https://github.com/scitran/core/issues/1072_

----

This feature had good intentions but current use of the system doesn't support the narrative of a normal user looking for other user's data and requesting access. Remove feature for now.

hkethi002:
Fixes #1072 

I don't know if I implemented `as_user` in the best way for the unit tests

### Review Checklist

### DIFF
--- a/api/handlers/dataexplorerhandler.py
+++ b/api/handlers/dataexplorerhandler.py
@@ -374,11 +374,12 @@ class DataExplorerHandler(base.RequestHandler):
             else:
                 modified_filters.append(f)
 
-        # Must be site admin to search all data
-        if request.get('all_data', False) and not self.user_is_admin:
-            raise APIPermissionException("Must have site admin privileges to search across all data")
-        # Add permissions filter to list if user is not requesting all data or is superuser
-        elif not request.get('all_data', False) and not self.superuser_request:
+        if request.get('all_data', False):
+            # User would like to search all data regardless of permissions
+            if not self.user_is_admin:
+                raise APIPermissionException("Must have site admin privileges to search across all data")
+        else:
+            # Add permissions filter to list if user is not requesting all data
             modified_filters.append({'term': {'permissions._id': self.uid}})
 
         # Only return objects that have not been marked as deleted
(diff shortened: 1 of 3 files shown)


api/handlers/dataexplorerhandler.py
```diff
@@ -370,8 +371,11 @@ def _parse_request(self, request_type='search'):
             else:
                 modified_filters.append(f)
 
+        # Must be site admin to search all data
+        if request.get('all_data', False) and not self.user_is_admin:
+            raise APIPermissionException("Must have site admin privileges to search across all data")
         # Add permissions filter to list if user is not requesting all data or is superuser
-        if not request.get('all_data', False) and not self.superuser_request:
+        elif not request.get('all_data', False) and not self.superuser_request:
             modified_filters.append({'term': {'permissions._id': self.uid}})
```
nagem: Thinking though the different scenarios to make sure this covered every one and to see if there was a more intuitive way to write it, I came up with the situations where a permissions query should be added:

|  | Non-Admin user | Admin user w/ root=false | Admin user w/ root=true
| ------------ | -------------|-------------- | ----------------------
| Would like to search just their data (all_data=false) | ✓ | ✓ |  - * This should be a check IMO
| Would like to search all data (all_data=true) | request denied | - | -

The UI is going to have to make a change no matter what, so I would suggest that "Orange Mode" users get the "All Data" checkbox selected for them by default, and site admins see the checkbox regardless of "mode". 

Then the if statement can be simplified to:
```
if request.get('all_data', False):
    # User would like to search all data regardless of permissions
    if not self.user_is_admin:
    	raise APIPermissionException("Must have site admin privileges to search across all data")
else:
  modified_filters.append({'term': {'permissions._id': self.uid}})
```

nagem:
This will require a UI change, so I added the "Breaking Change" tag.

nagem:
LGTM after we sort though the change I suggested.

AshleyAncona:
Ready to merge

PR #1183, opened 2018-05-02
Lab group flag

hkethi002:
Ref https://github.com/flywheel-io/team/issues/8
### Changes
- New `edition` property for groups
- List items an be `center` or `lab`

### Todo
- [X] ~~Delete all analyses in a group when toggled to center edition/ in session when the session is moved to a center~~ Prevent user from accessing analyses under center groups
- [x]  Add tests for analyses

### Review Checklist


api/auth/__init__.py
```diff
@@ -51,7 +51,7 @@ def check_login(self, *args, **kwargs):
 
 def require_admin(handler_method):
     """
-    A decorator to ensure the request is made as superuser.
+    A decorator to ensure the request is made as site admin.
```
nagem: 👌

api/web/base.py
```diff
@@ -297,6 +297,8 @@ def handle_exception(self, exception, debug, return_json=False): # pylint: disab
             custom_errors = exception.errors
         elif isinstance(exception, errors.APIUnknownUserException):
             code = 402
+        elif isinstance(exception, errors.APICenterEditionException):
+            code = 426  # Can't use 402 because that was logging me out
```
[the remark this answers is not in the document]
nagem: The "higher access level" would be a required for the object, not the user would be how I would spin that, but I see what you're saying. I still think `403` is the best option we have without using a generic `400`.

AshleyAncona:
Data migration plan: default group to lab, Thad run script for those that are not a lab over to center.

api/auth/groupauth.py
```diff
@@ -17,7 +17,7 @@ def f(method, _id=None, query=None, payload=None, projection=None):
                 handler.abort(403, 'not allowed to perform operation')
             elif _get_access(handler.uid, group) >= INTEGER_PERMISSIONS['admin']:
                 pass
-            elif method == 'GET' and _get_access(handler.uid, group) >= INTEGER_PERMISSIONS['ro']:
+            elif method == 'GET' and _get_access(handler.uid, group) >= INTEGER_PERMISSIONS['ro'] and "lab" not in payload:
```
nagem: Is this meant to be on a `GET`? They would not have a payload.

api/dao/hierarchy.py
```diff
@@ -78,6 +78,16 @@ def get_container(cont_name, _id):
...
     })
 
+def is_lab(cont_name, _id):
+    if cont_name == "collection":
+        return True
+    return get_parent_tree(cont_name, _id).get('group',{}).get('lab')
+
+def confirm_lab(cont_name, _id):
+    if not is_lab(cont_name, _id):
+        raise APICenterEditionException("Cannot create child for {} as a center.")
```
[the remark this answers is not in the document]
nagem: Even a simple `This action is reserved for groups marked as a lab.`

api/handlers/grouphandler.py
```diff
@@ -47,7 +47,7 @@ def handle_projects(self, result):
         return result
 
     def get_all(self, uid=None):
-        projection = {'label': 1, 'created': 1, 'modified': 1, 'permissions': 1, 'tags': 1}
+        projection = {'label': 1, 'created': 1, 'modified': 1, 'permissions': 1, 'tags': 1, 'lab': 1}
```
nagem: I wonder why we do this, and if it's still necessary.

nagem:
@ehlertjd thank you! If you have any final reviews please add them today, I'll give it a final review today as well. Hoping to merge on Monday.

kofalt:
Could we please use a string rather than a boolean? Suppose we decide to further segment the market.

api/auth/groupauth.py
```diff
@@ -15,7 +15,7 @@ def f(method, _id=None, query=None, payload=None, projection=None):
                 pass
             elif method in ['DELETE', 'POST']:
                 handler.abort(403, 'not allowed to perform operation')
-            elif _get_access(handler.uid, group) >= INTEGER_PERMISSIONS['admin']:
+            elif _get_access(handler.uid, group) >= INTEGER_PERMISSIONS['admin'] and "lab" not in payload:
```
kofalt: Won't this be incorrect if `lab` is present but set to false?

nagem:
>Could we please use a string rather than a boolean? Suppose we decide to further segment the market.

We discussed this on Monday, and came to the conclusion that we might have several "feature sets" that are added on to the base of "center". Might be easier long term to have an enum of available feature sets on each group.

kofalt:
I'd be fine with that, or a bitmap-ish string

api/auth/groupauth.py
nagem: The python `in` is just a key presence check, not a value check.

kofalt: That's what I'm getting at: if the `lab` key is present and false.

hkethi002:
@ehlertjd Instead of a boolean lab flag, it's now `edition` that can either be set to `center` or `lab`

AshleyAncona:
We want a list of values, rather than a string of values.  
The values will still be the edition.

api/dao/hierarchy.py
```diff
@@ -10,14 +10,18 @@
...
 from . import containerutil
 
 
 log = config.log
 
 PROJECTION_FIELDS = ['group', 'name', 'label', 'timestamp', 'permissions', 'public']
+EDITIONS = {
+    "center": 0,
+    "lab": 1
+}
```
ehlertjd: Is this used? If so, could you add a comment?

ehlertjd:
Would it be possible to update swagger documentation for endpoints to denote actions that are limited to lab edition? As a user, is there a good place I can look to see what the differences are between lab and center edition?

tests/unit_tests/python/test_dataexplorer.py
```diff
@@ -48,6 +48,15 @@ def test_search(as_public, as_drone, es, as_user):
...
                     {'range': filter_range},
+                    {'bool':
+                        {'should': [
+                            {'bool':
+                                {'must_not': [
+                                    {'term': {'container_type': 'analysis'}},
+                                    {'term': {'parent.type': 'analysis'}}
+                                ]}},
+                            {'terms': {'group.edition': 'lab'}}
+                        ]}},
```
ehlertjd: Maybe worth declaring this block as a constant and re-using it in tests.

nagem:
@hkethi002 there are no breaking changes in this PR as long as groups have `"lab"` in their `edition` list, correct?

hkethi002:
@nagem Correct, although new groups won't have a way to set it without a frontend change, (we're only setting it with the db upgrade)

nagem:
> new groups won't have a way to set it without a frontend change

I might be for defaulting a new groups that do not set the `edition` field themselves to have `["lab"]`. We can adjust this in a future PR if that is not a viable default.

nagem:
@AshleyAncona do you have any opinion on my default comment above?

nagem:
Did we decide to resurrect this PR?

hkethi002:
Making a new pull request

PR #1186, opened 2018-05-04
Ensure that j_id is always a string

ehlertjd:
With the switch to ObjectId in the devices table, joining acquisitions was failing due to keys being ObjectIds. This ensures that the join key is always a string.

### Review Checklist

### DIFF
--- a/api/handlers/containerhandler.py
+++ b/api/handlers/containerhandler.py
@@ -148,7 +148,7 @@ class ContainerHandler(base.RequestHandler):
 
             elif join_origin:
                 j_type = f['origin']['type']
-                j_id   = f['origin']['id']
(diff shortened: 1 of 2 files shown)


nagem:
I think this should also be fixed here, so origins always have string `id`s: https://github.com/flywheel-io/core/blob/master/api/web/base.py#L68

nagem:
There might be other code elsewhere that also assumes `origin.id` is a string, since it always was until the recent modification.

nagem:
It's not pretty but I saved `str()`ing all the origin fields until the end of the initialize function. I think that also fixed an issue we would have had [here](https://github.com/flywheel-io/core/pull/1186/files#diff-c86285960fea45f6471319c63cba0431R100)

PR #1188, opened 2018-05-07
Bootstrap fix

ambrussimon:
Fix test suite to enable re-running it multiple times against long-running API instance.
Show-stopping error was introduced via the `bootstrap` fixture that uses the new no-drone-secret-bootstrapping feature, but failed to clean up the `bootstrap` singleton from the DB thus tests could only run once.

After fixing, tried to run whole integration test suite twice, and found various similar smaller scoped cleanup errors (data leftovers after tests that result in errors on 2nd run). Started cleaning them up, but left some for the future...

Running subsets of tests multiple times is already functional, which was the original goal.

### Review Checklist

### DIFF
--- a/tests/integration_tests/python/conftest.py
+++ b/tests/integration_tests/python/conftest.py
@@ -374,6 +374,8 @@ class DataBuilder(object):
                 self.delete(child_cont, child['_id'], recursive=recursive)
         if resource == 'gear':
             _api_db.jobs.remove({'gear_id': str(_id)})
+        if resource == 'user':
+            _api_db.apikeys.delete_one({'uid': _id})
         _api_db[resource + 's'].remove({'_id': _id})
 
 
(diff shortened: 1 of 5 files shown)


tests/integration_tests/python/conftest.py
```diff
@@ -38,7 +38,8 @@ def bootstrap_users():
     data_builder = DataBuilder(session)
     data_builder.create_user(_id='user@user.com', api_key=SCITRAN_USER_API_KEY)
     yield data_builder
-    data_builder.teardown()
+    api_db.users.delete_many({})
```
ehlertjd: Maybe: `api_db.users.delete_many({'_id': {'$regex': r'@user\.com'}})`?

ehlertjd: Makes sense

PR #1189, opened 2018-05-07
Data views part 1

ehlertjd:
This first PR is a functional prototype. It does not yet support analyses or multiple files per container. There are also plans to support saved views, and a pre-canned set of columns.

I would like to get feedback on the general approach. Please keep in mind that I will be refactoring a good chunk of the DataView class into the [Pipeline Design Pattern](https://medium.com/@aaronweatherall/the-pipeline-pattern-for-fun-and-profit-9b5f43a98130) -- more for code extensibility and maintainability rather than to support parallel processing at this time.

The fundamental algorithm is unlikely to change significantly from this point forward.

### Review Checklist


nagem:
Haven't looked closely enough at the code to find this out for myself yet, but which container level are fair game for columns and filespecs in your design? Project/Subject/Session/Acquisition? Do we support Group columns? I know you mentioned something with Analyses and Files are coming, can I reference keys on those objects?

PR #1191, opened 2018-05-09
Pagination

### LINKED ISSUE #1187: Build Pagination Interface Proof of Concept
## Requirements:
  - Ability to sort on key chosen by the client
  - Ability to filter on exact match key-value pairs chosen by the client
  - Most functionality should be accessible via a `GET` request using query params, not a request that requires a request body
  - Pagination style should be `?after=<id>&limit=100` rather than `page=3&limit=100` or `start=300&limit=100`
  - All list endpoints should be able to (eventually) support this behavior
  - Must be opt-in

## Nice to haves:
  - Ability to filter using ranges (date/numeric)
  - Ability to sort on 0-many keys chosen by the client

## Considerations
 1. The optimal solution shouldn't have to be aware of which keys the requested container type has available for filtering and sorting, but if that is not possible due to type ambiguity it would be useful to have a standard way for container types to declare this information, and have it dynamically available in the documentation.

 1. It is possible `after` syntax is impossible in Mongo without doing a significant amount of work/at all. If that is the case we can drop it as a requirement. The justification for that syntax is that the you don't see redundant data on subsequent page requests if new data has entered the system since the last request. This would be especially troublesome for a resource like `jobs`, where new objects are created quite frequently. The limit of the `after` syntax is that jumping to page N is difficult.

ambrussimon:
Resolves #1187 

### Summary
__All*__ list endpoints now handle the following URL params:
* `?filter=expr[,expr...]` where `expr` is  `KEY OP VALUE`
   * example: `?filter=label=my-label,created>2018-09-22`
   * keys get into the query as-is
   * supported operators: <, <=, =, !=, =>, >
   * value casting: if it looks like an instance of _type_, then cast it (obj-id, datetime, float - in this order)
   * Note that `?filter=_id>X` is the current idiom for after_id
* `?sort=expr[,expr...]` where `expr` is `KEY [ : ORDER ]`
   * example: `?sort=label:asc,created:desc`
   * keys get into the query as-is
   * supported order values: `1`, `asc` -> ascending, `-1`, `desc` -> descending
   * default order if only key given: `asc`
* `?skip=N` where `N` is a positive integer
* `?page=N` where `N` is a positive integer (mutually exclusive with skip, requires limit)
* `?limit=N` where `N` is a positive integer
* Header `X-Accept-Feature: pagination` is a feature toggle for returning new-style list results
   * old: `[{"_id": "X"  "label": "my-acquisition", ...}, ...]`
   * new: `{"total": 42, "count": 10, "results": [{"_id": "X"  "label": "my-acquisition", ...}, ...]}`
* TODO swagger

*List endpoints (`GET`):
* `/devices` (new)
* `/users`
* `/groups`
* `/projects`
* `/sessions`
* `/acquisitions`
* `/analyses`
* `/collections`
* `/gears`
* `/site/rules`
* `/batch`
* and `/jobs` which is not yet exposed

### Review Checklist

### DIFF
--- a/api/dao/dbutil.py
+++ b/api/dao/dbutil.py
@@ -1,8 +1,10 @@
 import random
 import time
 
-from pymongo.errors import DuplicateKeyError
+import pymongo
+
 from ..web.errors import APIStorageException
+
 
 def try_replace_one(db, coll_name, query, update, upsert=False):
     """
@@ -14,7 +16,7 @@ def try_replace_one(db, coll_name, query, update, upsert=False):
 
     try:
         result = db[coll_name].replace_one(query, update, upsert=upsert)
-    except DuplicateKeyError:
+    except pymongo.errors.DuplicateKeyError:
         return result, False
     else:
         return result, True
@@ -39,3 +41,51 @@ def fault_tolerant_replace_one(db, coll_name, query, update, upsert=False):
             time.sleep(random.uniform(0.01,0.05))
 
     raise APIStorageException('Unable to replace object.')
+
+
+def paginate_find_kwargs(find_kwargs, pagination):
+    """
+    Modify and return find_kwargs (keyword arguments dict for `db.coll.find()`)
+    using the pagination settings.
+    """
+    if pagination:
+        if 'filter' in pagination:
+            filter_ = find_kwargs.get('filter', {})
+            filter_.update(pagination['filter'])
+            find_kwargs['filter'] = filter_
+
+        if 'sort' in pagination:
+            sort = find_kwargs.get('sort', [])
+            if isinstance(sort, basestring):
+                sort = [(sort, pymongo.ASCENDING)]
+            sort.extend(pagination['sort'])
+            find_kwargs['sort'] = sort
+
+        if 'limit' in pagination:
+            find_kwargs['limit'] = pagination['limit']
+    return find_kwargs
+
+
+def paginate_pipeline(pipeline, pagination):
+    """
+    Modify and return mongo pipeline (list of stages for `db.coll.aggregate()`)
+    using the pagination settings.
+    """
+    if pagination:
+        if 'pipe_key' in pagination:
+            pipe_key = pagination.pop('pipe_key')
+            for key in pagination.get('filter', {}).keys():
+                pagination['filter'][pipe_key(key)] = pagination['filter'].pop(key)
+            for i, key_order in enumerate(pagination.get('sort', [])):
+                key, order = key_order
+                pagination['sort'][i] = (pipe_key(key), order)
+
+        if 'filter' in pagination:
+            pipeline.append({'$match': pagination['filter']})
+
+        if 'sort' in pagination:
+            pipeline.append({'$sort': pagination['sort']})
+
+        if 'limit' in pagination:
+            pipeline.append({'$limit': pagination['limit']})
+    return pipeline
--- a/api/dao/dbutil.py
+++ b/api/dao/dbutil.py
@@ -98,7 +98,8 @@ def paginate_pipe(collection, pipeline, pagination):
 
     # total_pipeline.append({'$count': 'total'})  # mongo 3.4+
     total_pipeline.append({'$group': {'_id': None, 'total': {'$sum': 1}}})
+    total_result = list(collection.aggregate(total_pipeline))
     return {
-        'total': next(collection.aggregate(total_pipeline))['total'],
+        'total': total_result[0]['total'] if total_result else 0,
         'results': list(collection.aggregate(pipeline)),
     }
--- a/api/dao/dbutil.py
+++ b/api/dao/dbutil.py
@@ -66,7 +66,7 @@ def paginate_find(collection, find_kwargs, pagination):
             find_kwargs['limit'] = pagination['limit']
 
     return {
-        'total': collection.count(find_kwargs['filter']),
+        'total': collection.count(find_kwargs.get('filter', {})),
         'results': list(collection.find(**find_kwargs)),
     }
 
--- a/api/dao/dbutil.py
+++ b/api/dao/dbutil.py
@@ -84,7 +84,7 @@ def paginate_pipeline(pipeline, pagination):
             pipeline.append({'$match': pagination['filter']})
 
         if 'sort' in pagination:
-            pipeline.append({'$sort': pagination['sort']})
+            pipeline.append({'$sort': dict(pagination['sort'])})
 
         if 'limit' in pagination:
             pipeline.append({'$limit': pagination['limit']})
--- a/api/dao/dbutil.py
+++ b/api/dao/dbutil.py
@@ -44,11 +44,8 @@ def fault_tolerant_replace_one(db, coll_name, query, update, upsert=False):
     raise APIStorageException('Unable to replace object.')
 
 
-def paginate_find_kwargs(find_kwargs, pagination):
-    """
-    Modify and return find_kwargs (keyword arguments dict for `db.coll.find()`)
-    using the pagination settings.
-    """
+def paginate_find(collection, find_kwargs, pagination):
+    """Return paginated `db.coll.find()` results."""
     if pagination:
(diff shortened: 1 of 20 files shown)


nagem:
For the UI (and possibly other clients as well), we're going to need a way to express how many items are in the collection that satisfy the filters - sorry that wasn't in the original writeup request. The only way I know of to do that is to wrap the actual results in a `results` key and return other top level information like `count`. Is this something we would want to do every time the client requests paginated results or would a no context list of results be the default and only return the results under a key if the client specifically asked for the count?

api/util.py
```diff
@@ -128,6 +129,27 @@ def is_group_id(gid):
...
+        r'^\d\d\d\d-\d\d-\d\d$':                              '%Y-%m-%d',
+        r'^\d\d\d\d-\d\d-\d\dT\d\d$':                         '%Y-%m-%dT%H',
+        r'^\d\d\d\d-\d\d-\d\dT\d\d:\d\d$':                    '%Y-%m-%dT%H:%M',
+        r'^\d\d\d\d-\d\d-\d\dT\d\d:\d\d:\d\d$':               '%Y-%m-%dT%H:%M:%S',
+        r'^\d\d\d\d-\d\d-\d\dT\d\d:\d\d:\d\d\.\d\d\d\d\d\d$': '%Y-%m-%dT%H:%M:%S.%f',
+    }
+    for date_re, date_fmt in re_fmt.iteritems():
+        if re.match(date_re, s):
+            return datetime.datetime.strptime(s, date_fmt)
+    return None
```
nagem: I'm still not entirely sure we don't want to just lock it down to ISO format only. Specifically with the first regex, any 4 digit number will get caught in that. An example: `jobs/?filter=context.runtime_in_milliseconds>1000`

I'm not sure what that actual jobs key is, but it does exist.

api/util.py
```diff
@@ -128,6 +129,24 @@ def is_group_id(gid):
...
     return bool(pattern.match(gid))
 
+
+def datetime_from_str(s):
+    """
+    Return datetime.datetime parsed from a string that is a prefix of isoformat.
+    Return None if the string is not such a valid prefix.
+    """
+    re_fmt = {
+        r'^\d\d\d\d-\d\d-\d\d$':                              '%Y-%m-%d',
```
ehlertjd: `\d{4}` might be slightly nicer to read?

api/web/base.py
```diff
@@ -286,6 +287,100 @@ def is_true(self, param):
...
+
+            try:
+                pagination['limit'] = int(limit_param)
+                if pagination['limit'] <= 0:
+                    raise ValueError('expected positive integer, got {}'.format(pagination['limit']))
+            except ValueError as e:
+                msg = 'Invalid pagination query param "limit": {}'
+                raise errors.APIValidationException({'error': msg.format(e.args[0])})
+
+        return pagination
```
ehlertjd: How do I select the next "page"? `offset=N`?

api/web/base.py
```diff
@@ -286,6 +287,100 @@ def is_true(self, param):
...
+            filter_ops = {'<':  '$lt',
+                          '<=': '$lte',
+                          '=':  '$eq',
+                          '!=': '$ne',
+                          '>=': '$gte',
+                          '>':  '$gt'}
+            for filter_str in filter_param.split(','):
+                for filter_op in sorted(filter_ops, key=len, reverse=True):
+                    key, op, value = filter_str.partition(filter_op)
+                    if op:
```
ehlertjd: Maybe break this logic out into a separate function for the production version?

api/dao/dbutil.py
```diff
@@ -39,3 +41,51 @@ def fault_tolerant_replace_one(db, coll_name, query, update, upsert=False):
...
+                pagination['filter'][pipe_key(key)] = pagination['filter'].pop(key)
+            for i, key_order in enumerate(pagination.get('sort', [])):
+                key, order = key_order
+                pagination['sort'][i] = (pipe_key(key), order)
+
+        if 'filter' in pagination:
+            pipeline.append({'$match': pagination['filter']})
+
+        if 'sort' in pagination:
+            pipeline.append({'$sort': dict(pagination['sort'])})
```
ehlertjd: I think this should be a `collections.OrderedDict`

api/util.py
ambrussimon: This was easier on the eyes for me, TBH. If you feel strong about it and/or others also think `{4}` is nicer, just bump this and will do.

api/web/base.py
ambrussimon: Using the `after_id` idiom the usual way of getting
* the 1st page is: `?limit=N`
* any other page: `?limit=N&after_id=123`

`?after_id=123` is equivalent to `?filter=_id>123`, and for now only the latter is supported. Happy to add `after_id` as a synonym for that.

Also, `?offset=N` is very straightforward to implement (and is also performant, since `skip=N` has mongo support), so I was thinking pagination could support _both_ paging styles.

api/dao/dbutil.py
ambrussimon: Truly nice catch.

nagem:
Notes from offline team discussion:

  - `filter=_id>X` would only work if the primary (only) sort was by `_id`. A few other options have been raised, will open a new ticket to track that work. Worth releasing with `limit`+`skip`
  - API will use new return format (results list under `results` key, additional information at top level dict) when client requests it using an X-Header
  - `count` should be renamed to something that clarifies it is the count of all objects that match the filter, not the number returned by that request. Possibly `total_results` or just `total`.
  - Support `page=<N>&limit=<N2>` where the API will do the required math to compute the `skip`. If both `page` and `skip` are present, `400`

kofalt:
Will the consumer have to know ahead of time to use `_id` or `id`?

kofalt:
I would support `total` in addition to `count`, the later of which is the # of results returned in that call.

api/handlers/containerhandler.py
```diff
@@ -329,7 +329,8 @@ def get_all(self, cont_name, par_cont_name=None, par_id=None):
         else:
             query = {}
         # this request executes the actual reqeust filtering containers based on the user permissions
-        results = permchecker(self.storage.exec_op)('GET', query=query, public=self.public_request, projection=projection)
+        page = permchecker(self.storage.exec_op)('GET', query=query, public=self.public_request, projection=projection, pagination=self.pagination)
+        results = page['results']
         if results is None:
             self.abort(404, 'No elements found in container {}'.format(self.storage.cont_name))
```
nagem: I'm not sure I agree with this existing code, and it would be even more confusing with filters to get a 404 if your filters don't match anything while you are tuning them. Anyone disagree with removing it?

nagem:
@ambrussimon great turnaround on those requests, as always. I took a look through and changes look good. I'll give it one more in depth passthrough soon - if there is anything in the review you would like me to spend more time on, let me know. 

Is the plan to allow access to the `/jobs` endpoint in a separate PR? Is that much additional work?  

Also: 

>Handle filters on PHI-related fields/logging. For now use blacklisted PHI fields stored on ContainerStorage class to deny requests that attempt to filter on those fields.

As to the final bullet above, I'm talking with Ryan and Erik on how we should handle that situation. With the current implementation in this PR, we are in line with Search's interpretation of how that access should be logged. I'll let you know soon if we should make any changes.

api/handlers/containerhandler.py
ehlertjd: @nagem I agree with the policy of list endpoints returning an empty result rather than 404.

nagem:
Other than the 404 change above I'd say feel free to merge. If client (UI/SDK) development work finds anything that would require a change, that can be in a separate followup PR. Great work, don't forget to open a PR with the `/jobs` endpoint exposed as well.

ambrussimon:
Addressed everything but swaggering. @ehlertjd if there's any chance that this can be templated somehow, could you please give me some pointers? If not, I imagine the implementation to be tedious (1d+), large in diff and ugly.

PR #1199, opened 2018-05-11
Data views part 3

ehlertjd:
This is the final set of API changes for data views. This adds the ability to save views under the user, group or project containers, or a global `"site"` container. In addition, it provides a queryable mapping of predefined column aliases.

### Review Checklist


nagem:
I might have missed it, but where do we filter/check user input for requested columns? What would happen if a user attempted to retrieve the column `acquisitions.files`?

PR #1211, opened 2018-05-22
Fix tests for storage

ambrussimon:
For E2E-testing flywheel-io/storage-plugins#8, I updated fly/fly's `docker/run-tests.sh` to use core+, and I ran into some errors (unrelated to the referenced plugins PR) and fixed them.

Note that storage-plugins installs a newer pylint in core+, which reveals a bunch of errors. Here I only fixed the minor/obvious little lint problems.

The long-term plan is to upgrade pylint to the same version and fix remaining errors to make the images more coherent and to enable core+ integration testing via fly/fly using non-local storage.

### Review Checklist

### DIFF
--- a/tests/integration_tests/python/test_download.py
+++ b/tests/integration_tests/python/test_download.py
@@ -261,10 +261,9 @@ def test_filelist_range_download(data_builder, as_admin, file_form):
     assert r.ok
     ticket = r.json()['ticket']
 
-    # try to download single file's first byte w/o feature flag
+    # verify contents
     r = as_admin.get(session_files + '/one.csv',
-                     params={'ticket': ticket},
-                     headers={'Range': 'bytes=0-0'})
+                     params={'ticket': ticket})
     assert r.ok
     assert r.content == '123456789'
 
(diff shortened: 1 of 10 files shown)


tests/integration_tests/python/test_download.py
```diff
@@ -411,20 +410,21 @@ def test_filelist_range_download(data_builder, as_admin, file_form):
...
-    boundary = r.headers.get('Content-Type').split('boundary=')[1]
-    assert r.content == '--{0}\n' \
-                        'Content-Type: text/csv\n' \
-                        'Content-Range: bytes 1-2/9\n\n' \
-                        '23\n' \
-                        '--{0}\n' \
-                        'Content-Type: text/csv\n' \
-                        'Content-Range: bytes 3-4/9\n\n' \
-                        '45\n'.format(boundary)
+    if os.getenv('SCITRAN_PERSISTENT_FS_URL', 'osfs').startswith('osfs'):
```
nagem: Why was this changed to only test multiple ranges when using osfs? Is it not supported with other file systems?

nagem:
One small question otherwise LGTM!

tests/integration_tests/python/test_download.py
ambrussimon: Correct, while testing with GS, I got a 400 with an XML body stating that multiple ranges are not supported.

PR #1213, opened 2018-05-22
Project-level gear visibility

kofalt:
Adds a `projects` key to `GearDoc` that affects `/gears` visibility. No other routes are affected.

With `dcm2niix` containing `projects: [ "5" ]` and `dicom-mr-classifier` containing `projects: [ ]`:

```
$ 'api/gears' | jq '.[].gear.name'

"flywheel-example-gear"
"feature-check"
"dicom-mr-classifier"
"qa-report-fmri"
"afq-demo"
```

```
$ 'api/gears?project=5' | jq '.[].gear.name'

"flywheel-example-gear"
"feature-check"
"dicom-mr-classifier"
"dcm2niix"
"qa-report-fmri"
"afq-demo"
```

```
$ 'api/gears?project=6' | jq '.[].gear.name'

"flywheel-example-gear"
"feature-check"
"dicom-mr-classifier"
"qa-report-fmri"
"afq-demo"
```

It's possible that the `projects` filtering could find its way into `dbutil.paginate_find` syntax instead; patches accepted.

### DIFF
--- a/api/jobs/gears.py
+++ b/api/jobs/gears.py
@@ -18,23 +18,45 @@ from ..web.errors import APIValidationException, APINotFoundException
 
 log = config.log
 
-def get_gears(pagination=None):
+def get_gears(project=None, pagination=None):
     """
-    Fetch the install-global gears from the database
+    Fetch the install-global gears from the database. Filters out project-hidden gears.
     """
 
     if pagination:
         pagination['pipe_key'] = lambda key: 'original.' + key
 
+    # Find gears that are not hidden by project
+    hide_gears = {
+        '$match': {
+            '$or': [
+                { 'projects': { '$exists': False }},
+                { 'projects': { '$eq': [] }}
+            ]
+        }
+    }
+
+    if project:
+        # Add in gears that match the current project
+        hide_gears['$match']['$or'].append({
+            'projects': {
+                '$in': [ project ]
+            }
+        })
+
     pipe = [
-        {'$sort': {
-            'gear.name': 1,
-            'created': -1,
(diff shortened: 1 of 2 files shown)


ehlertjd:
I was a little surprised that we don't validate against an input schema for gear_doc - is that something that we should address in a future PR?

kofalt:
> Will I ever want to be able to add another project to the projects subdoc?

Yes, I think we'll want an endpoint for that in this PR.

> Can I ever remove a project from the list?

Fair question, I hadn't thought of it but it would be easy to add.

> Should gears that belong to projects get deleted if the project referring to them is deleted?

Probably not. The manifest and gear asset are important and irreplaceable. Maybe a `/gears/orphans` or similar would be worthwhile? Combine with a new admin dashboard component?

> we don't validate against an input schema for gear_doc - is that something that we should address in a future PR?

Totally - the jobs module has historically been a little more wildcat. It would need to reference the current `gears` module, maybe, or just punt on the actual "gear" part since that is explicitly validated.

ehlertjd:
Where I was going with my line of questioning is - I don't think we want you to remove projects (or at least not all projects) otherwise the gear will then become public on the instance.

kofalt:
Sure, though that's also the way to mark the gear as public to the site (which is an eventuality we should support). I feel like "remove last project" is something that can just be a client-side warning of the effect.

kofalt:
Holding due to priority change

nagem:
Closing for now, can be reopened if the need/priority winds change again. Leaving branch intact.

PR #1224, opened 2018-05-25
Add 3-way storage file access and migration tests

ambrussimon:
Merged functionality early to fix master, tests added just now.
Includes tests for v1 files:
* core serves them
* migrate_storage moves them 
* core continues to serve them

### Review Checklist


nagem:
This seems like a reasonable way to handle it. We should discuss long term plans with @ryansanford / @tcbtcb to see if we can eventually migrate everyone to either v1 or remote storage and remove the double legacy path logic.

PR #1226, opened 2018-05-29
Cleanup deleted files

davidfarkas:
### Review Checklist


nagem:
Initial pass looks good, but I believe we would also want to remove deleted containers from the db which complicates the origin:user thing a bit. We should discuss with @ambrussimon.

nagem:
@davidfarkas This can be merged when you are ready.

PR #1228, opened 2018-05-29
Extract measurements from Custom classification

ehlertjd:
Per PM: 
We don't want the prior file measurements values showing in the UI, so rather than store them in the `Custom` classification array, we should keep them in the old `measurements` field.

This PR consists of:
- A database upgrade to extract the original measurements values from the 'Custom' field
- Logic to unset `measurements` if the file classification or modality ever changes
- Logic to prevent returning the `measurements` field from endpoints

### Review Checklist

### DIFF
--- a/api/dao/basecontainerstorage.py
+++ b/api/dao/basecontainerstorage.py
@@ -93,6 +93,13 @@ class ContainerStorage(object):
             parent_name = child_name
         return parent_tree
 
+    @classmethod
+    def filter_container_files(cls, cont):
+        if cont is not None and cont.get('files', []):
+            cont['files'] = [f for f in cont['files'] if 'deleted' not in f]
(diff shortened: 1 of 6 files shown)


ryansanford:
@ehlertjd 
Is there concern with cutting a release without this change? Should this PR be labeled Breaking Change?

gsfr:
@ryansanford Yes, this is a release blocker.

@ehlertjd Please clarify:
> Logic to prevent returning the `measurements` field from endpoints

Are `measurements` only handed to gears?

ehlertjd:
@gsfr Yes, measurements are only handed to gears.

nagem:
Changes look correct for the request, I'll catch up with @ehlertjd later today to get more details about the request itself. 

LGTM

PR #1230, opened 2018-05-31
Performance refactor on GET /gears and /sessions/x/jobs

### LINKED ISSUE #1227: Calls to /jobs and /gears have significantly slowed
If you take a look at https://dev.flywheel.io/branch/master/#/dashboard/session/576be45259b2b6346719ed71?tab=data you'll notice that the file names do not appear until ~1 second after loading the page.  We can fix that on the frontend, but the root cause is the calls to /jobs and /gears are taking almost a second to return.

ambrussimon:
Closes #1227 
TLDR; potentially not solving actual problem

* improved pagination performance (/gears)
   * detail: getting the total _AND_ the (filtered/limited) results is non-trivial
   * used to do 2 separate queries, now avoiding that with more advanced mongo-fu
* refactored `/sessions/x/jobs` - now using significantly less mongo queries

Used `cProfile` and _synthetic data_:
```
gear.name:a-z, gear.version:0-9 (260 gears total)
5(analysis+job) and 5(acquisition+job) PER gear_id (5200 jobs total)

time (s) spent in handler according to cProf:

           GET /gears      GET /sessions/x/jobs?join=gears
(master)   0.006           3.589
(pr)       0.003           1.587
```

**However**, testing via the UI showed that actual E2E performance was not improved that much, the requests from the issue are easily getting to ~1s still. I need further input if this doesn't fully solve the problem. My next idea would be to reduce response size by trimming anything not explicitly used, but maybe others can chime in with better approaches.

### Review Checklist

### DIFF
--- a/api/dao/dbutil.py
+++ b/api/dao/dbutil.py
@@ -66,10 +66,11 @@ def paginate_find(collection, find_kwargs, pagination):
             find_kwargs['limit'] = pagination['limit']
 
     results = collection.find(**find_kwargs)
-    return {
+    page = {
         'total': results.count(),  # count ignores limit and skip by default
         'results': list(results),
     }
+    return page
 
 
 def paginate_pipe(collection, pipeline, pagination):
@@ -104,5 +105,4 @@ def paginate_pipe(collection, pipeline, pagination):
             pipeline.append({'$project': {'total': 1, 'results': {'$slice': ['$results'] + slice_args}}})
 
     page = next(collection.aggregate(pipeline), {'total': 0, 'results': []})
-    page['results'] = list(page['results'])
     return page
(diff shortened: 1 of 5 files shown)


api/dao/dbutil.py
```diff
@@ -85,21 +85,24 @@ def paginate_pipe(collection, pipeline, pagination):
...
+        elif 'limit' in pagination:
+            slice_args = [pagination['limit']]
+        elif 'skip' in pagination:
+            # NOTE aggregation + pagination.skip without limit not supported
+            pass
+        if slice_args:
+            pipeline.append({'$project': {'total': 1, 'results': {'$slice': ['$results'] + slice_args}}})
+
+    page = next(collection.aggregate(pipeline), {'total': 0, 'results': []})
+    page['results'] = list(page['results'])
```
ehlertjd: What was `results`, if not a list already?

ambrussimon: Thanks, removed unnecessary cast.

PR #1233, opened 2018-06-01
Fix multi-file download

### LINKED ISSUE #1232: Multi-file downloads hanging on object storage
Behavior at stanfordlabs instance, which has object storage enabled:
* Multi-file downloads, such as one entire acquisition or files from multiple acquisitions, hang for extended periods with no user-feedback before data starts coming in, as well as what appears to be before each large file.
* If one of the files is *too large*, the downloaded tar file is incomplete and cannot be opened with Mac Finder.
* If the first file is *too large*, the Flywheel UI disappears and a non-user-friendly error message is shown.
```
{"status_code": 504, "message": "Nginx: gateway timeout."}
```

This behavior represents a significant deterioration of the user experience and should be considered a bug as well as blocker to deploying object storage.

cc @AshleyAncona @ryansanford @lmperry

davidfarkas:
* Resolves #1232
* Requires tandem storage-plugins PR flywheel-io/storage-plugins#18 (merged to master)

#### Problem
* when adding a member to a tarfile, `TarFile.gettarinfo(fileobj=member_fd)` is called
* that calls `member_fd.fileno()` for `os.stat()`-ing
* as a graceful fallback for all-around compatibility, storage files downloaded the whole thing to temp, and yielded that fileno
* thus before even starting to stream contents, the whole thing was downloaded

#### Solution
* instead of using (or monkeypatching) `TarFile.gettarinfo()` for storage files, provide custom `TarInfo`

### Review Checklist

### DIFF
--- a/api/download.py
+++ b/api/download.py
@@ -283,8 +283,11 @@ class Download(base.RequestHandler):
         with tarfile.open(mode='w|', fileobj=stream) as archive:
             for filepath, arcpath, cont_name, cont_id, _ in ticket['target']:
                 file_system = files.get_fs_by_file_path(filepath)
-                with file_system.open(filepath, 'rb') as fd:
-                    yield archive.gettarinfo(fileobj=fd, arcname=arcpath).tobuf()
+                with file_system.openbin(filepath, 'rb') as fd:
+                    if hasattr(fd, 'gettarinfo'):
+                        yield fd.gettarinfo(arcname=arcpath).tobuf()
+                    else:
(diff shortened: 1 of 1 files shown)


ambrussimon:
@ehlertjd I think that's likely. Couldn't quite get it to work yet, but will definitely clean it up and open separate PR if indeed possible to use existing `getinfo`.

davidfarkas:
@ehlertjd If the info that `getinfo` can provide is enough (size, type, mtime) then yes we can use that. The reason why I left the original code in place because in case of OS file system it provides a more detailed (gid, uid, linkname, uname, gname)`TarInfo` instance.

PR #1238, opened 2018-06-05
Increase the download speed of multi-file downloads

davidfarkas:
With this solution I was able to reduce the number of download requests to one request during archive download which is the root cause of the slow download speed.
It should significantly increase the archive download performance. I wasn't able to check what is the maximum downloads speed because I reached the office network capacity which is 
6 MB/s.

### Review Checklist

### DIFF
--- a/api/download.py
+++ b/api/download.py
@@ -7,6 +7,8 @@ import cStringIO
 
 import fs.path
 import fs.errors
+from fs.iotools import RawWrapper
+import requests
 
 from .web import base
 from .web.request import AccessType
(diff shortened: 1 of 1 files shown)


api/download.py
```diff
@@ -283,18 +285,25 @@ def archivestream(self, ticket):
         with tarfile.open(mode='w|', fileobj=stream) as archive:
             for filepath, arcpath, cont_name, cont_id, _ in ticket['target']:
                 file_system = files.get_fs_by_file_path(filepath)
-                with file_system.openbin(filepath, 'rb') as fd:
-                    if hasattr(fd, 'gettarinfo'):
-                        yield fd.gettarinfo(arcname=arcpath).tobuf()
+                with file_system.open(filepath, 'rb') as fd:
+                    if isinstance(fd, RawWrapper) and hasattr(fd._f, 'gettarinfo'):  # pylint: disable=protected-access
```
ehlertjd: Are there any filesystems (existing or planned) that implement RawWrapper but don't have gettarinfo (or vice-a-versa)?

api/download.py
```diff
@@ -283,18 +285,25 @@ def archivestream(self, ticket):
...
+                        yield fd._f.gettarinfo(arcname=arcpath).tobuf()  # pylint: disable=protected-access
+
+                        signed_url = files.get_signed_url(filepath, file_system)
+
+                        response = requests.get(signed_url, stream=True)
```
ehlertjd: So if I understand this correctly, the root cause was that pyfilesystem was opening >1 request for the signed url when using the chunked `fd.read`, and we're using the signed url directly to avoid that? Is that something that can/should be fixed in the filesystem implementation, or is it unavoidable?

Does `file_system.open` still open a second request? Is that one avoidable?

If `read` is this inefficient, can the zip info and zip member endpoints be improved with similar changes?

api/download.py
```diff
@@ -283,18 +285,25 @@ def archivestream(self, ticket):
...
+                            yield chunk
                     else:
                         yield archive.gettarinfo(fileobj=fd, arcname=arcpath).tobuf()
-                    chunk = ''
-                    for chunk in iter(lambda: fd.read(CHUNKSIZE), ''): # pylint: disable=cell-var-from-loop
-                        yield chunk
-                    if len(chunk) % BLOCKSIZE != 0:
-                        yield (BLOCKSIZE - (len(chunk) % BLOCKSIZE)) * b'\0'
+                        chunk = ''
+                        for chunk in iter(lambda: fd.read(CHUNKSIZE*15), ''):  # pylint: disable=cell-var-from-loop
```
ehlertjd: Is the intent to leave this streaming in 15MB chunks? If so, can we document why in a comment here?

davidfarkas: It's a mistake, I will set it back. Thanks for the heads up

api/download.py
davidfarkas: For example OSFS doesn't implement RawWrapper

api/download.py
ambrussimon: > So if I understand this correctly, the root cause was that pyfilesystem was opening >1 request for the signed url when using the chunked fd.read,

Yes, the root cause was `1 chunk read == 1 range read request`

> and we're using the signed url directly to avoid that?

Yes, the above solution is effectively one request for the whole file, and now it starts yielding from the response before getting it all.

> Is that something that can/should be fixed in the filesystem implementation, or is it unavoidable?

The filesystem behaves like that to enable a _seekable_ file like object interface without having to download the whole file, to facilitate (now proven too slow for UI) random reads.
This solution (`response.iter_content`) for this endpoint is only applicable _because_ we need the whole file, it's bytes in order, and we do not `seek`.

> Does file_system.open still open a second request? Is that one avoidable?

Yes, `.open()` incurs an extra request, one that checks whether the file exists or not. Only avoidable if it's OK to `raise` late, when trying to `read()`

> If read is this inefficient, can the zip info and zip member endpoints be improved with similar changes?

Not without caching and/or implementing an in-house `ZipFile`, that optimizes all the `seek`s for object storage

ehlertjd:
The place where I still see long delays is creating the download ticket. This results in seemingly nothing happening for 2-15s after clicking download in the UI.

For example, this ticket was for 8 files on my local system, using object storage:
screen shot 2018-06-05 at 9 30 39 am

ehlertjd:
And this one is for 6 files on dev:
screen shot 2018-06-05 at 9 35 18 am

davidfarkas:
Yep, because we check that the file exists or not during creating the ticket.
```
file_path, _ = files.get_valid_file(file_obj)
            if file_path:  # silently skip missing files
                targets.append((file_path, cont_name+'/'+cont_id+'/'+file_obj['name'], cont_name, cont_id, file_obj['size']))
                total_size += file_obj['size']
                file_cnt += 1
```
Possible solution is to move this check to the `archivestream` method.

davidfarkas:
https://github.com/flywheel-io/storage-plugins/pull/23 improves the performance of creating the download ticket.

PR #1243, opened 2018-06-06
Port range read and signed urls for analyses files

davidfarkas:
The reason why analyses files weren't be redirected, because `/analyses` enpoint uses the `AnalysesHandler` and signed url redirection was only implemented in the `FileListHandler`. This is my fault, I didn't know that it uses a different handler. 
I ported the code from the `FileListHandler` because it is less risky and quicker than putting it to a common place. Also we will rewrite or refactor these handlers soon when we move the files into its own collection.

### Review Checklist

### DIFF
--- a/api/handlers/refererhandler.py
+++ b/api/handlers/refererhandler.py
@@ -405,6 +405,9 @@ class AnalysesHandler(RefererHandler):
 
                 # Request to download the file itself
                 else:
+                    # START of duplicated code
+                    # IMPORTANT: If you modify the below code reflect the code changes in
+                    # listhandler.py:FileListHandler's download method
                     signed_url = files.get_signed_url(file_path, file_system,
                                                       filename=filename,
                                                       attachment=(not self.is_true('view')),
@@ -481,6 +484,7 @@ class AnalysesHandler(RefererHandler):
                                         self.response.write('\n')
                                     else:
                                         self.response.write(data)
+            # END of duplicated code
 
             # log download if we haven't already for this ticket
             if ticket:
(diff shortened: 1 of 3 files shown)


nagem:
@ehlertjd - @davidfarkas is right, the next thing on the list is tearing these two copied handlers apart and creating a central handler class anyway. There are a ton of existing calls to `self` for response writing and reading so putting the combined code anywhere other than on a common parent handler or passing access to the response object around would be tricky. I suggested the lower risk but more messy solution of a direct copy since the source is already almost a direct copy. I do understand your concern, though.

@davidfarkas is there some amount of this code that could be placed in a central location? Do the integration tests written give us any more confidence?

api/handlers/refererhandler.py
```diff
@@ -404,13 +405,82 @@ def download(self, **kwargs):
...
+                            self.response.headers['Content-Length'] = str(
+                                fileinfo['size'])  # must be set after setting app_iter
+
+                            if self.is_true('view'):
+                                self.response.headers['Content-Type'] = str(
+                                    fileinfo.get('mimetype', 'application/octet-stream'))
+                            else:
+                                self.response.headers['Content-Type'] = 'application/octet-stream'
+                                self.response.headers['Content-Disposition'] = 'attachment; filename="' \
+                                                                               + str(filename) + '"'
```
ehlertjd: Should [listhandler.py:478](https://github.com/flywheel-io/core/blob/master/api/handlers/listhandler.py#L478) also cast `filename` to `str`?

davidfarkas: Somehow the `filename` was unicode only here. I can cast that there too if you think it is more safer.

ehlertjd: Without knowing the root cause, and with the idea that these code-paths will be merged one day, I think I'd make the change.

davidfarkas:
Great, once the CI finishes, I will merge it.

PR #1245, opened 2018-06-07
Place .MISSING placehorder files in the tar if not exists

davidfarkas:
To decrease the time of creating the download ticket we only check that a file exists or not during creating the archive. If it doesn't then we place an empty file which ends with `.MISSING` to indicate that we missed that.

### Review Checklist

### DIFF
--- a/api/download.py
+++ b/api/download.py
@@ -288,29 +288,41 @@ class Download(base.RequestHandler):
             for filepath, arcpath, cont_name, cont_id, _ in ticket['target']:
                 try:
                     file_system = files.get_fs_by_file_path(filepath)
-                except fs.errors.ResourceNotFound:
-                    tarinfo = TarInfo()
-                    tarinfo.name = arcpath + '.MISSING'
-                    yield tarinfo.tobuf()
-                else:
                     with file_system.open(filepath, 'rb') as fd:
-                        if isinstance(fd, RawWrapper) and hasattr(fd._f, 'gettarinfo'):  # pylint: disable=protected-access
+                        signed_url = None
+                        if isinstance(fd, RawWrapper) and hasattr(fd._f,  # pylint: disable=protected-access
+                                                                  'gettarinfo'):
                             tarinfo = fd._f.gettarinfo(arcname=arcpath).tobuf()  # pylint: disable=protected-access
                             yield tarinfo
                             signed_url = files.get_signed_url(filepath, file_system)
-
-                            response = requests.get(signed_url, stream=True)
-                            for chunk in response.iter_content(chunk_size=CHUNKSIZE):
-                                yield chunk
-                                if len(chunk) % BLOCKSIZE != 0:
-                                    yield (BLOCKSIZE - (len(chunk) % BLOCKSIZE)) * b'\0'
                         else:
                             yield archive.gettarinfo(fileobj=fd, arcname=arcpath).tobuf()
+
+                        try:
+                            if signed_url:
+                                response = requests.get(signed_url, stream=True)
+                                f_iter = response.iter_content(chunk_size=CHUNKSIZE)
+                            else:
+                                f_iter = iter(lambda: fd.read(CHUNKSIZE), '')  # pylint: disable=cell-var-from-loop
+
                             chunk = ''
-                            for chunk in iter(lambda: fd.read(CHUNKSIZE), ''):  # pylint: disable=cell-var-from-loop
+                            for chunk in f_iter:
                                 yield chunk
                             if len(chunk) % BLOCKSIZE != 0:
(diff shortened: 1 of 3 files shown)


api/download.py
```diff
@@ -284,23 +286,31 @@ def archivestream(self, ticket):
...
-                        for chunk in iter(lambda: fd.read(CHUNKSIZE), ''):  # pylint: disable=cell-var-from-loop
-                            yield chunk
-                        if len(chunk) % BLOCKSIZE != 0:
-                            yield (BLOCKSIZE - (len(chunk) % BLOCKSIZE)) * b'\0'
+                try:
+                    file_system = files.get_fs_by_file_path(filepath)
+                except fs.errors.ResourceNotFound:
+                    tarinfo = TarInfo()
+                    tarinfo.name = arcpath + '.MISSING'
+                    yield tarinfo.tobuf()
```
nagem: We'll want to make sure we log a `critical` level warning here, and would like to eventually log the information to prometheus for better monitoring capabilities. Information that would seem relevant is parent cont type, parent cont id, file name, and then file path or hash or uuid. Path as well as hash/uuid might be interesting to know if our path logic was off.

api/download.py
```diff
@@ -284,23 +286,31 @@ def archivestream(self, ticket):
...
+                else:
+                    with file_system.open(filepath, 'rb') as fd:
```
ehlertjd: Should we also be catching IOError on `file_system.open` and produce a `.MISSING` file? If we don't catch the exception here, we're probably going to truncate the download.

nagem:
By doing a manual DB modification of a file you should be able to write a test for this fairly easily.

api/download.py
```diff
@@ -284,23 +286,31 @@ def archivestream(self, ticket):
...
+                        if isinstance(fd, RawWrapper) and hasattr(fd._f, 'gettarinfo'):  # pylint: disable=protected-access
+                            tarinfo = fd._f.gettarinfo(arcname=arcpath).tobuf()  # pylint: disable=protected-access
+                            yield tarinfo
+                            signed_url = files.get_signed_url(filepath, file_system)
+
+                            response = requests.get(signed_url, stream=True)
+                            for chunk in response.iter_content(chunk_size=CHUNKSIZE):
+                                yield chunk
+                                if len(chunk) % BLOCKSIZE != 0:
+                                    yield (BLOCKSIZE - (len(chunk) % BLOCKSIZE)) * b'\0'
```
davidfarkas: This fixes a bug which was introduced with my previous fix. The tar was broken when we downloaded more than one file, and can't be extracted. If I opened the archive I only saw the first file.

api/download.py
```diff
@@ -284,23 +286,31 @@ def archivestream(self, ticket):
...
+                    with file_system.open(filepath, 'rb') as fd:
+                        if isinstance(fd, RawWrapper) and hasattr(fd._f, 'gettarinfo'):  # pylint: disable=protected-access
+                            tarinfo = fd._f.gettarinfo(arcname=arcpath).tobuf()  # pylint: disable=protected-access
+                            yield tarinfo
+                            signed_url = files.get_signed_url(filepath, file_system)
+
+                            response = requests.get(signed_url, stream=True)
+                            for chunk in response.iter_content(chunk_size=CHUNKSIZE):
+                                yield chunk
+                                if len(chunk) % BLOCKSIZE != 0:
```
ehlertjd: I think this if-block should be de-dented, we shouldn't be yielding padding until after the last block was sent.

davidfarkas: I think the condition will be true only when we are actually at the last chunk, otherwise `len(cunck)` will be exactly the `CHUNKSIZE`.

api/download.py
davidfarkas: Yep, that makes sense

api/download.py
ehlertjd: http://docs.python-requests.org/en/master/api/#requests.Response.iter_content

> Iterates over the response data. When stream=True is set on the request, this avoids reading the content at once into memory for large responses. The chunk size is the number of bytes it should read into memory. **This is not necessarily the length of each item returned as decoding can take place.**

api/download.py
davidfarkas: Hmmm, and what if a `read` causes IOError?

ehlertjd: My take: 

If we've already sent the the file header there's not much we can do. I would think abort the transfer and log a critical message. 

If we haven't sent the file header and we encounter an error, then I think we can replace it with the `.MISSING` entry (and log a critical message)

api/download.py
davidfarkas: Got it

api/download.py
davidfarkas: That was my idea too after I sent the comment, so great, I will do this.

nagem:
Tests look good, nice work 👍

PR #1247, opened 2018-06-07
Update URLs in job's request after retry

nagem:
Fix is complete. Retried jobs were getting a copy of the previous job's request block, but the inputs that referenced a URI with the previous job id in them were not getting updated. This fixes that, as well as a bug where a job with an `_id` before insert was getting it's `_id` dropped and replaced. Confusing `id` vs `_id` vs `id_` will need to be dealt with eventually. 

### Review Checklist

### DIFF
--- a/api/jobs/jobs.py
+++ b/api/jobs/jobs.py
@@ -217,7 +217,7 @@ class Job(object):
     def mongo(self):
         d = self.map()
         if d.get('id'):
-            d['id'] = bson.ObjectId(d['id'])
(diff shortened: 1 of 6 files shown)


api/jobs/jobs.py
```diff
@@ -217,7 +217,7 @@ def map(self):
     def mongo(self):
         d = self.map()
         if d.get('id'):
-            d['id'] = bson.ObjectId(d['id'])
+            d['_id'] = bson.ObjectId(d.pop('id'))
```
kofalt: Hahahaha, um

Issue #1248, opened 2018-06-08
/api/jobs/<jobid>/retry fails with 500


ryansanford:
Found while testing the tip of master (commit 45bdbd1cddc369badca876b60331c38ee10dfd3a)

@kofalt I wonder if this is related to change #1247 ?

```
2018-06-08 16:40:15      scitran.api                  base.py   414:DEBU from ryansanford@flywheel.io POST /api/jobs/5b1ab060e417a7001c1dad72/retry {} request_id=9a78a6f4-1528476015
2018-06-08 16:40:15      scitran.api                  base.py   393:ERRO Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/webapp2.py", line 570, in dispatch
    return method(*args, **kwargs)
  File "/var/scitran/code/api/api/jobs/handlers.py", line 561, in retry
    for x in j.inputs:
TypeError: 'NoneType' object is not iterable
 request_id=9a78a6f4-1528476015
```

kofalt:
Looks like a job without inputs cannot be retried. Easy fix

ehlertjd:
Closed via #1250

PR #1250, opened 2018-06-08
Check for Job.inputs == None on retry handler

kofalt:
Quick fix for #1248

### DIFF
--- a/api/jobs/handlers.py
+++ b/api/jobs/handlers.py
@@ -557,10 +557,13 @@ class JobHandler(base.RequestHandler):
         j = Job.get(_id)
 
         # Permission check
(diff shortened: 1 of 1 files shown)


api/jobs/handlers.py
```diff
@@ -557,7 +557,7 @@ def retry(self, _id):
         j = Job.get(_id)
 
         # Permission check
-        if not self.superuser_request:
+        if not self.superuser_request and j.inputs is not None:
             for x in j.inputs:
```
hkethi002: This may skip the access check on the destination when there are no inputs

kofalt:
Nice catch. Should be fixed!

PR #1253, opened 2018-06-13
Fix UID re-upload 404 on deleted containers

### LINKED ISSUE #1241: Reaper upload fails when there is matching uid in deleted session
Route: /api/upload/reaper
Current response: HTTP 404

Expected behavior: The deleted object(s) should be replaced. This enables iterative reaper upload testing with the same source data.

Suspect other upload routes are impacted, but haven't confirmed.

ambrussimon:
Fixes #1241 

All UID-based upload endpoints were affected and are now fixed.
Warns when "prematurely" purging something marked as deleted from mongo like this.

### Update / PR rename

Thanks @hkethi002, I realized the implicit deletion is wrong/unneeded. The problem was `_create_query()` missing `{"deleted": {"$exists": False}}`.

### Review Checklist

### DIFF
--- a/api/dao/hierarchy.py
+++ b/api/dao/hierarchy.py
@@ -241,12 +241,12 @@ def upsert_fileinfo(cont_name, _id, fileinfo, access_logger):
     container_after, file_before = None, None
 
     for f in container_before.get('files',[]):
-        # Fine file in result and set to file_after
+        # Find file in result and set to file_after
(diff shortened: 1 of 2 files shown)


ambrussimon:
If we ever _really_ want to not implicitly delete files either, I realized one could rename `file.name` to `file.deleted_name` or something, so it could still be theoretically recovered before the final cleanup.

PR #1254, opened 2018-06-13
Remove permissions from analyses

hkethi002:
### Changes
- Downloading analysis nodes uses parent permissions
- Permissions list is never created
- DB upgrade to remove current permissions lists (Most likely they are all out of date anyways as the propagate changes function doesn't propagate to analyses)

### Review Checklist

### DIFF
--- a/api/download.py
+++ b/api/download.py
@@ -1,5 +1,6 @@
 import os
 import bson
+import copy
 import pytz
 import tarfile
 import datetime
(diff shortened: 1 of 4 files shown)


api/download.py
```diff
@@ -219,8 +219,14 @@ def _preflight_archivestream(self, req_spec, collection=None):
                 total_size, file_cnt = self._append_targets(targets, 'acquisitions', acq, prefix, total_size, file_cnt, req_spec.get('filters'))
 
             elif item['level'] == 'analysis':
+                perm_query = base_query.pop('permissions._id')
                 analysis = config.db.analyses.find_one(base_query, ['parent', 'label', 'inputs', 'files', 'uid', 'timestamp'])
-                if not analysis:
+                base_query["permissions._id"] = perm_query
```
nagem: Will need to use a copy of the base query to modify. With this format, the next node in the for loop will no longer have permissions as part of the query.

hkethi002: I think the query is being reset right there but it'd probably be safer in general to make a copy

nagem: Oh you are right, I missed the line that reset it.

PR #1256, opened 2018-06-19
Add Prometheus Metrics

ehlertjd:
This currently collects the following metrics from the uwsgi instance:
1. **Request:** method, path, response code, response time and bytes sent
2. **Workers:** Process CPU and memory usage
3. **System:** CPU, Memory, I/O and disk usage
4. **Elastic:** Whether or not it's running
5. **Database:** Version, Release, Job counts by state and some collection counts

I have not touched most of the rest of the codebase, but it would probably be a good idea to increment a counter for every `log.critical` call.

Fairly easy to add/remove stuff in any of the general categories.

Metrics are exposed via `/api/metrics` -- which is not currently authenticated, which **MUST** be remediated before merging. Possible options:
1. Store metrics on persistent storage rather than publish an endpoint
2. Add prometheus access key or use drone key

### Review Checklist


ryansanford:
We can also block requests to `/api/metrics` from outside. What would be an appropriate response for that approach? a 404?

nagem:
Ah thanks for the clarification, @ambrussimon. I forget the drone key is the old shared secret, not a device key. I meant to agree with the device key option.

api/metrics/mule.py
```diff
@@ -0,0 +1,179 @@
...
+        info = config.es.info()
+        values.ELASTIC_IS_UP.set(1)
+
+        version = info.get('version', {})
+        build_hash = version.get('build_hash', '')
+        lucene_version = version.get('lucene_version', '')
+        number = version.get('number', '')
+
+        # Capture elastic version
+        values.ELASTIC_VERSION.labels(build_hash, lucene_version, number).set(1)
```
nagem: I know we are gather some amount of elastic metrics somewhere, @ryansanford do you know where I can find those?

ehlertjd: @nagem It appears that we're using the [elasticsearch-exporter](https://github.com/justwatchcom/elasticsearch_exporter)

It's arguable whether or not exporting anything about elastic from the uwsgi container is useful, I actually added those before I realized that we were collecting some metrics elsewhere!

api/metrics/mule.py
```diff
@@ -0,0 +1,179 @@
...
+    workers = {}
+
+    while True:
+        with values.COLLECT_METRICS_TIME.time():
+            collect_system_metrics()
+            collect_worker_metrics(workers)
+            collect_elastic_metrics()
+            collect_db_metrics()
+
+        time.sleep(POLL_INTERVAL)
```
kofalt: The problem with this is that the collection & presentation of metrics is now completely separated from the polling interval, which is prometheus' responsibility.

Instead, how about using mule messages to notify the mule that metrics have been retrieved? Refresh on message. Now metrics are one-poll-interval out of date, rather than an arbitrary amount, which would allow prometheus to adjust its monitoring pressure accordingly.

api/metrics/mule.py
```diff
@@ -0,0 +1,179 @@
...
+    """
+    try:
+        import uwsgi
+        # For now, just cleanup dead workers
+        dead_workers = set(workers.keys())
+        for worker in uwsgi.workers():
+            pid = worker['pid']
+
+            if pid in dead_workers:
+                dead_workers.remove(pid)
```
kofalt: Why have this?

api/metrics/request_wrapper.py
```diff
@@ -0,0 +1,106 @@
...
+        try:
+            response_time = max(default_timer() - self._start, 0)
+
+            template = 'UNKNOWN'
+            if hasattr(self._request, 'route') and hasattr(self._request.route, 'template'):
+                template = self._request.route.template
+
+            method = getattr(self._request, 'method', 'UNKNOWN')
+
+            labels = [ method, template, str(self._status) ]
```
kofalt: Uhh... won't this result in a huge set of labels?

Maybe that's a good idea. Dunno...

api/metrics/values.py
```diff
@@ -0,0 +1,63 @@
...
+SYSTEM_NETWORK_BYTES_SENT = Gauge('system_network_bytes_sent', 'Observed network bytes sent', [], multiprocess_mode='max')
+SYSTEM_NETWORK_BYTES_RECEIVED = Gauge('system_network_bytes_received', 'Observed network bytes received', [], multiprocess_mode='max')
+# Disk I/O
+SYSTEM_DISK_READ_COUNT = Gauge('system_disk_read_count', 'Observed disk read count', [], multiprocess_mode='max')
+SYSTEM_DISK_WRITE_COUNT = Gauge('system_disk_write_count', 'Observed disk write count', [], multiprocess_mode='max')
+SYSTEM_DISK_READ_BYTES = Gauge('system_disk_read_bytes', 'Observed disk read bytes', [], multiprocess_mode='max')
+SYSTEM_DISK_WRITE_BYTES = Gauge('system_disk_write_bytes', 'Observed disk write bytes', [], multiprocess_mode='max')
+# Disk Usage
+SYSTEM_DISK_BYTES_USED = Gauge('system_disk_bytes_used', 'Observed disk usage, in bytes', ['path'], multiprocess_mode='max')
+SYSTEM_DISK_BYTES_FREE = Gauge('system_disk_bytes_free', 'Observed disk availability, in bytes', ['path'], multiprocess_mode='max')
```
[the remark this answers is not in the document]
ehlertjd: Unfortunately [None](https://github.com/prometheus/client_python#multiprocess-mode-gunicorn):
> Custom collectors do not work (e.g. cpu and memory metrics)

The custom collectors work by pulling at collection time, which the multiprocess mode doesn't do, so we're stuck with a push model.

Open to alternatives, but thats how I understand it anyway.

api/metrics/request_wrapper.py
ehlertjd: You do get one label per template string in `api/api.py`, but if you don't separate them out, then you have endpoints like `/api/dataexplorer/index/fields` skewing the average.

api/metrics/mule.py
ehlertjd: Yeah, if we're certain that we want to pull the metrics from the `/api/metrics` endpoint (which it seems like we are) then this shouldn't take too much to wire up.

kofalt:
In addition to the mule doc link, it'd probably make sense to record this information:

https://github.com/prometheus/client_python#multiprocess-mode-gunicorn

api/metrics/mule.py
```diff
@@ -0,0 +1,179 @@
...
+
+        # Get raw collection counts
+        for collection_name in COUNT_COLLECTIONS:
+            count = config.db[collection_name].count()
+            values.COLLECTION_COUNT.labels(collection_name).set(count)
+
+    except: # pylint: disable=bare-except
+        log.exception('Error collecting db metrics')
+
+def run_mule():
```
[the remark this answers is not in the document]
ehlertjd: That is correct

api/metrics/mule.py
kofalt: What I'm getting at is, was nothing removing dead workers before? Do you foresee side effects of doing so here?

api/metrics/values.py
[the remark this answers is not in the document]
kofalt: Yeah, I think I ran into that right after finishing the review in my bottom comment :) This makes sense

api/metrics/mule.py
ehlertjd: This is really just the metrics view of workers, doesn't affect the rest of the system.

api/metrics/request_wrapper.py
kofalt: Maybe we can try it this way and see how "bad" it gets. By contrast, the engine clumps I/O and non I/O requests and leaves it at that. I'm concerned / interested to see what this will do: each route, each customer, each poll. We can always try it out!

kofalt:
LGTM pending a shift to mule messages / explicit choice to stay on our own interval.

api/metrics/request_wrapper.py
ehlertjd: One possible alternative would be to label by handler class+method, should be a slightly smaller surface area, I'm open to suggestions - the template strings could use some reformatting to make them readable, should we continue to use them.

kofalt: If we can have "all the labels" it would be useful, I just don't know where it becomes a problem for our Prometheus instance. I mentioned this to Ryan, I think he'll have a good idea of "how many is too many".

kofalt: I like the idea of maybe just class, if we have to reduce. That alone is probably enough information to diagnose problems.

kofalt:
Fair point. I just want to be able to query a site without drama - if I can at least hit it from outside its container when SSHed locally that's enough for me. Maybe no auth for now & block remote with nginx.

api/metrics/request_wrapper.py
ryansanford: The more dangerous thing is having a label value that changes over time for the same intended metric. The current implementation doesn't seem to be doing that.

We definitely will have metric explosion with one value per routing template string, but the value/payoff seems high as well. I'll do a little work to estimate the storage cost. The happy medium may be a longer polling interval for that metric set.

I will ask that `uwsgi_response_time_seconds_count` and `uwsgi_response_size_bytes_count` get combined into a singular `uwsgi_response_count` metric. The labels for that metric must match `uwsgi_response_time_seconds_count` and `uwsgi_response_size_bytes_sum`

ehlertjd:
@ryansanford the response counts have been combined into `uwsgi_response_count`

api/metrics/values.py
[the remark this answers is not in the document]
ryansanford: @ehlertjd If the system metrics are truly system scoped, I'd prefer that concern stay with node-exporter (https://github.com/prometheus/node_exporter) which is already present in the Flywheel Dev and Prod environments.

ryansanford:
I'd prefer the metrics to be prefixed with the service or component that they are specific to. As as result, I suggest...
- `worker` prefix change to `uwsgi_worker`
- `uwsgi` prefix change to `fw_core` or `fw_api`

api/metrics/values.py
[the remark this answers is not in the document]
ehlertjd: @ryansanford The "system" metrics are local to the uwsgi container. I believe that the metrics that I've added will be useful in troubleshooting performance issues in the future, but they could potentially be replaced with container metrics, if capturing those is feasible.

ehlertjd:
@ryansanford I believe that all of your comments have been addressed

kofalt:
Ryan brought up a really good point, I put everything under `fw_engine` for sanity's sake.

Strongly recommend `fw_core` over `_api`!

kofalt:
(Sorry... I know...)

PR #1262, opened 2018-06-22
Implement after_id

ambrussimon:
* Added `after_id=x` style paging
* Cannot be used together with sort, page, skip (APIValidationError)
* Does not work with `/gears` (as it's aggregation relies on sort, it returns 500 with after_id)

### Review Checklist

### DIFF
--- a/api/dao/dbutil.py
+++ b/api/dao/dbutil.py
@@ -49,7 +49,11 @@ class PaginationError(Exception):
 
 
 def paginate_find(collection, find_kwargs, pagination):
-    """Return paginated `db.coll.find()` results."""
+    """Return paginated `db.coll.find()` results.
+
+    Raises PaginationError if the query is incompatible with the pagination:
+     * `sort` in find_kwargs and `after_id` in pagination
+    """
     if pagination:
         if 'after_id' in pagination:
             if find_kwargs.get('sort'):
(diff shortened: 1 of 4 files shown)


api/dao/dbutil.py
```diff
@@ -44,9 +44,19 @@ def fault_tolerant_replace_one(db, coll_name, query, update, upsert=False):
     raise APIStorageException('Unable to replace object.')
 
 
+class PaginationError(Exception):
```
ehlertjd: Maybe I missed this - will this raise a `500` and is that the intended behavior?

hkethi002: Also, I think the error class should be in in `web/errors.py`

nagem: If you take a look at https://github.com/flywheel-io/core/pull/1262/files#diff-c86285960fea45f6471319c63cba0431R327 all input should be validated by the time this is raised - it might be an error reserved for when a developer uses the `paginate_find` or `paginate_pipe` without properly validating input. There might be a clearer way to write this or adding a stipulation to the function comment block might help.

nagem:
LGTM other than addressing and finding a resolution to the exception confusion.

api/dao/dbutil.py
ambrussimon: To clarify:
* @ehlertjd yes, `PaginationError` raises an intended 500
* @nagem it's raised when trying to apply a valid pagination (`{"after_id": 1}`) on a **query** that's not compatible (`find_kwargs={"sort": {"name": 1}}`)
* example: `/gears?after_id=x` - the gear aggregation query relies on sorting

Added docs.

PR #1263, opened 2018-06-22
Enable storage testing

ambrussimon:
* coreplus building on top of core upgraded some python packages
* instead of pinning old deps in coreplus, I upgraded core requirements
* one of those is pylint, hence the code changes
* changes will be needed for successful fly/fly integration testing (== `docker/run-tests.sh` == CI) after flywheel-io/flywheel#634 lands

### Review Checklist


api/config.py
```diff
@@ -118,8 +120,7 @@ def apply_env_variables(config):
     if not os.path.exists(_path):
         os.makedirs(_path)
     __config['persistent']['fs_url'] = 'osfs://' + _path
-
-log.setLevel(getattr(logging, __config['core']['log_level'].upper()))
+log.debug('Persistent fs url: %s', __config['persistent']['fs_url'])
```
nagem: Is this something you want to keep long term or a temporary testing log?

api/dao/hierarchy.py
```diff
@@ -581,22 +580,3 @@ def _update_container_nulls(base_query, update, container_type):
...
-        parsed = parsed_files.get(info['name'])
-        if parsed:
-            path = parsed.path
-            new_infos = copy.deepcopy(parsed.info)
-        else:
-            path = None
-            new_infos = {}
-        new_infos.update(info)
-        merged_files[info['name']] = files.ParsedFile(new_infos, path)
-    return merged_files
```
nagem: 👍

api/handlers/dataexplorerhandler.py
```diff
@@ -547,7 +547,7 @@ def get_nodes(self):
 
     def get_file_nodes(self, return_type, filters, search_string):
 
-        query = self._construct_file_query(return_type, filters, search_string)['query']
+        query = self._construct_query(return_type, filters, search_string)['query']
```
nagem: Are you sure this is a correct change? @hkethi002 can you take a look?

nagem: Oh I think it's supposed to be `_construct_exact_query()`

api/config.py
ambrussimon: It is intended for testing/debugging only. It already proved useful to me, and it's also in line with other persistence vars (data_path & mongo_uri) being printed when in debug level.

(On the long term, we might want to avoid logging at import-time, but that's a different issue.)

api/handlers/dataexplorerhandler.py
ambrussimon: * `get_file_nodes()` is only called once, and with `return_type='file'`
* `_construct_query(return_type='file',...)` calls `_construct_exact_query()`
* the above is the only place `construct_exact_query()` is called
* thus it seemed like `construct_exact_query()` is... "more private", so I used the other one

ambrussimon:
@nagem This is once again ready for review, with questions answered and rebase fixed (metrics lint).

api/handlers/dataexplorerhandler.py
nagem: Thanks for chasing that down, current choice sounds good to me. `DataExplorer` could use a reorg after initial pass...

requirements.txt
```diff
@@ -3,19 +3,19 @@ django>=1.11.0,<1.12.0
...
 jsonschema==2.6.0
-Markdown==2.6.5
+Markdown==2.6.11
 prometheus_client==0.2.0
 psutil==5.4.6
-pymongo==3.2
-python-dateutil==2.4.2
-pytz==2015.7
-requests[security]==2.18.4
+pymongo==3.6.1
```
nagem: Looks like most of these are minor version changes. I confirmed compatibility with the pymongo lib. Let me know if you think any update deserves a closer inspection.

PR #1264, opened 2018-06-25
[FLYW-51] User _id switch to bson.ObjectId

hkethi002:
### Changes
- User id is an ObjectId
- handler.uid is an ObjectId
- origin.id for is still a string to match the ids of other origin.types
- permissions use ObjectId
- Collection curator is ObjectId
- `_id` is not a valid input for user, `email` is required for posts (Frontend change needed)
- when viewing access_logs that are of origin.type=="user", the frontend will need to display the origin.email field instead of _id
### Todo
- [x] DB update
- [ ] Figure out Frontend changes
### Review Checklist

### DIFF
--- a/swagger/schemas/definitions/job.json
+++ b/swagger/schemas/definitions/job.json
@@ -61,6 +61,7 @@
         "type": {
           "type": "string"
         },
+        "email":  {"$ref": "user.json#/definitions/email"},
         "id": {
           "type": ["string", "null"]
         }
--- a/api/api.py
+++ b/api/api.py
@@ -34,8 +34,8 @@ routing_regexes = {
     # Group ID: 2-32 characters of form [0-9a-z.@_-]. Start and ends with alphanum.
     'gid': '[0-9a-z][0-9a-z.@_-]{0,30}[0-9a-z]',
 
-    # User ID: any length, [0-9a-z.@_-]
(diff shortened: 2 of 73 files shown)


hkethi002:
Access logs will have the string of the ObjectId in the origin, so when the logs are returned, we can't tell who they logs are attributed to. We could add in the email to the origin, so that jobs and files would also have access to user emails.

swagger/schemas/definitions/job.json
```diff
@@ -58,6 +58,7 @@
         "type": {
           "type": "string"
         },
+        "email":  {"$ref": "user.json#/definitions/email"},
```
kofalt: Why this field?

kofalt:
Why are we moving towards ObjectIds? We previously agreed we wanted to move away from these, what changed?

swagger/schemas/definitions/job.json
kofalt: Ugh, okay. Not a fan, I'd rather see us app-join them when appropriate, but oh well

ehlertjd: Now that a user's email address is mutable, I think you would (*probably *?) want the address at the time the log message was captured, rather than the current email address.

PR #1266, opened 2018-06-26
Framework for observing container events

ehlertjd:
This decouples the logic to update a container from actions taken when a container is updated.

This is not quite complete - if the approach is approved, I will also update the appropriate locations in `placer.py` to notify when containers are updated.

### Review Checklist

### DIFF
--- a/api/dao/containereventhandlers.py
+++ b/api/dao/containereventhandlers.py
@@ -1,0 +1,54 @@
+from .. import config
+from ..dao.containerevents import register_container_observer
+
+log = config.log
+
+def register():
+    """Register global event handlers"""
+    # Create jobs when files are updated (any container)
+    register_container_observer(create_jobs_event_handler, event_types=['container_files_updated'])
+
+    register_container_observer(update_session_compliance_handler_for_files, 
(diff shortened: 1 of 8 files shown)


nagem:
I wanted to give this the initial 👍. I like the direction it's heading. I have a few questions I'll post soon but the overall idea is interesting and I think very useful for a few upcoming tasks we have on the docket.

kofalt:
It doesn't look like events can have data attached to them, would that be possible?

api/dao/containereventhandlers.py
```diff
@@ -0,0 +1,56 @@
...
+    from ..jobs.rules import create_jobs
+    if not event.container_before:
+        # Don't try to create jobs if we don't have a "before" container
+        if isinstance(result, dict):
+            result['jobs_spawned'] = 0
+        return
+
+    jobs_spawned = create_jobs(config.db, event.container_before, event.container_after, event.container_type)
+    result['jobs_spawned'] = len(jobs_spawned)
+
```
nagem: Noting an offline conversation about how these two functions below need to be separated. Is it because you already have the container available to you in `update_session_compliance_handler_for_acquisition`? Might make sense to put the check in one combined function.

PR #1268, opened 2018-06-26
Add context log to Request and RequestHandler

ehlertjd:
Replaces the default `config.log` with a [LoggerAdapter](https://docs.python.org/2/library/logging.html#loggeradapter-objects) that can be updated with arbitrary key-value pairs.

Also adds a `log` object to `RequestHandler` that includes the request origin in the context. And updates all 8 locations where we were logging from RequestHandlers.

Future PR may attempt to do some additional logging cleanup

For example:
```python
log = config.log.with_context(job_id=result.inserted_id)
```

### Review Checklist


nagem:
I like this format, SGTM

PR #1270, opened 2018-06-27
Improve API Exception classes

### LINKED ISSUE #690: Make schema validation errors a superset of standard errors
<a href="https://github.com/kofalt"></img></a> **Issue by [kofalt](https://github.com/kofalt)**
_Tuesday Mar 14, 2017 at 18:09 GMT_
_Originally opened as https://github.com/scitran/core/issues/690_

----

We have a few spots where we return client-parsable information about a validation error:

```python
return {
	'reason': 'Gear manifest does not match schema',
	'error': err.message.replace("u'", "'"),
	'key': key
}
```

The issue is that this does not overlap with the standard api error struct:

```python
{
	"message": "Something went wrong",
	"status_code": 500
}
```

It is hugely convenient if all errors are a superset of the same base struct.

I propose we keep message useful, and add the extra fields like so:

```python
message = err.message.replace("u'", "'")

return {
	'message':  'Gear manifest does not match schema on key ' + key ': ' + message,
	'status_code': 400,

	'reason': 'Gear manifest does not match schema',
	'error': message,
	'key': key
}
```

This keeps `message` a useful all-in-one, while still breaking out the client-friendly keys afterwards.

ehlertjd:
There is now a single base APIException which carries a `status_code`, `core_status_code`, and `default_message`. 

Validation errors include additional client-parseable attributes (closes #690)

Wherever possible, I eliminated the following anti-pattern:
```python
except APISomeException as e:
    self.abort(400, e.message)
```

I did **NOT** try to replace all instances of `self.abort` with `raise APIException` -- I think that should be a separate PR.

Refs #807 

### Review Checklist

### DIFF
--- a/api/dao/containerstorage.py
+++ b/api/dao/containerstorage.py
@@ -499,7 +499,7 @@ class AnalysisStorage(ContainerStorage):
         job = analysis.pop('job', None)
         if job is not None:
             if parent_type != 'session':
-                raise APIValidationException({'reason': 'Analysis created via a job must be at the session level'})
+                raise APIValidationException(reason='Analysis created via a job must be at the session level')
             analysis.setdefault('inputs', [])
             for key, fileref_dict in job['inputs'].iteritems():
                 analysis['inputs'].append(fileref_dict)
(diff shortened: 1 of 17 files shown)


kofalt:
Omg yessss

ehlertjd:
@hkethi002 Nice catch - totally unused exception, so I removed it.

PR #1271, opened 2018-06-27
Add gear metadata to jobs and analyses

### LINKED ISSUE #1190: Denormalize some gear details on analyses
For #1189 Data Views - It would be great/essential to be able to filter analyses based on some gear details (e.g. id or name+version) in order to perform bulk file retrieval.

I propose adding those three fields at a minimum to analyses, with migration script.

ehlertjd:
Add the following fields to jobs and analyses on creation (and upgrade):
- `gear_name`
- `gear_version`
- `gear_category`

In addition, add the `gear_id` field to analysis.

I'm guessing it will be useful to be able to search some of these fields, but I haven't added them to the connector yet.

Closes #1190

### Review Checklist


hkethi002:
It might just be easier and more sane to include it, but won't the gear category be "analysis" for every analysis?

ehlertjd:
@hkethi002  Today for sure. Don't know if that will stay the case forever, if someone else is confident that it will, it's pretty easily removed.

kofalt:
It would be nice to isolate these keys under an object, which makes it a little clearer that it's a partial-join cache thing. Or, to put it another way, I will sigh sadly if we don't, haha. But no super-strong opinion

ehlertjd:
@kofalt Done, I chose `gear_info` as the subdoc name

ehlertjd:
@nagem I would agree with that - I was hesitant to call it `gear` on the subdoc because it's a super subset of the gear manifest.

nagem:
>I'm guessing it will be useful to be able to search some of these fields, but I haven't added them to the connector yet.

They won't need to be added explicitly unless we want to change the format.

bin/database.py
```diff
@@ -1726,6 +1726,41 @@ def upgrade_to_51():
...
     config.db.analyses.update_many({}, {"$unset": {"permissions": ""}})
+
+def upgrade_job_to_52(job, gears):
+    job_id = job.get('_id')
+    gear_id = str(job.get('gear_id'))
+    gear_doc = gears.get(gear_id)
+
+    if not gear_doc:
+        logging.warn('Unable to upgrade job {} to 52 - gear {} does not exist!'.format(job['_id'], gear_id))
+        return True
```
nagem: Would we want to still save some form of an empty `gear_info` map? Or use the concept of the `dummy_gear` from `upgrade_to_22()`? Probably not, but I wanted to raise the option.

ehlertjd: @nagem I guess I'd be inclined not to worry about it unless you can think of any value it would add.

Issue #1273, opened 2018-06-28
Code coverage sometimes fails to merge


ehlertjd:
Example build where this happens:
https://travis-ci.org/flywheel-io/core/jobs/397942669

Which results in an (hopefully!) incorrect coverage report.

PR #1276, opened 2018-07-03
Disable codecov until build/coverage problems can be addressed

nagem:
Ref #1273


nagem:
FYI @ehlertjd @hkethi002 @ambrussimon @davidfarkas @kofalt

PR #1284, opened 2018-07-10
Analysis improvements

### LINKED ISSUE #908: Analysis Improvements
_Thursday Aug 24, 2017 at 21:35 GMT_
_Originally opened as https://github.com/scitran/core/issues/908_

----

  - [x] Split analysis into `files` (outputs) and `inputs`. Inputs are a reference to files on another container and should be secondary to files only attached to the analysis 
 
 - [x] Keep `job=true` logic for `POST api/<container_name>/<container_id>/analyses`, move default behavior to new endpoint, possibly `POST api/<container_name>/<container_id>/analyses/legacy`

 - [x] Make job inflation optional when requesting an analysis/many analyses

 - [x] Add routes to access analyses directly. Leave existing routes for now (?)
    - `analyses/<id>`
    - `analyses/<id>/files`
    - `analyses/<id>/files/<filename>`
    - `analyses/<id>/notes`
    - `analyses/<id>/notes/<id>`

- [ ] Remove old analyses endpoints and maybe move around some of the other endpoints to be more intuitive. May cause breaking changes for frontend.

Known endpoints to change:

/sessions/{SessionId}/analyses/{AnalysisId}/files
/sessions/{SessionId}/analyses/{AnalysisId}/files/{FileName}

Note: Keeping breaking changes minimal to the "old" analysis upload type is important while also phasing out/redesigning that use case.

### LINKED ISSUE #1242: Analysis joins
Implement missing analysis joins:
- `?join=origin`
- `?join=origin_job_gear_name`

#1208 was closed via #1217 - `?join_avatars=true`

ambrussimon:
Resolves #908 with notable exception of removing redundant endpoints to avoid breaking.
Resolves #1242 

Team, it was a real thrill to use the exception classes. <3

### Review Checklist

### DIFF
--- a/api/dao/basecontainerstorage.py
+++ b/api/dao/basecontainerstorage.py
@@ -9,6 +9,7 @@ from . import dbutil
 from .. import config
 from .. import util
 
+from ..types import Origin
 from ..web.errors import APIStorageException, APIConflictException, APINotFoundException
 
 log = config.log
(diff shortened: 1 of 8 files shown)


api/dao/basecontainerstorage.py
```diff
@@ -414,3 +415,56 @@ def join_avatars(containers):
...
+                    j_types = containerutil.pluralize(j_type)
+                    join_doc = config.db[j_types].find_one({'_id': j_id_b})
+
+                    # Join in gear name on the job doc if requested
+                    if join_doc is not None and join_gear_name and j_type == 'job':
+
+                        gear_id = join_doc['gear_id']
+                        gear_name = None
+
+                        if gear_id not in cached_gears:
```
ehlertjd: Does #1271 allow us to cut down on additional queries?

api/handlers/listhandler.py
```diff
@@ -486,7 +486,7 @@ def get(self, cont_name, list_name, **kwargs):
 
                     for first, last in ranges:
                         if first > fileinfo['size'] - 1:
-                            self.abort(416, 'Invalid range')
+                            raise RangeNotSatisfiable()
```
ehlertjd: 🎉

api/dao/basecontainerstorage.py
ambrussimon: It does indeed, thanks! Since we always have access to gear_name for free now, I made `?join=origin_job_gear_name` a noop, it's just always included.

PR #1287, opened 2018-07-12
Carry redundant node_type on resolver node

ehlertjd:
This PR does 2 things:
- Carry `node_type` as well as `container_type` on resolver nodes
- Allow discriminating resolver node on either `node_type` or `container_type` in the SDK

This will allow prior clients to use the new resolver, and will allow new clients to use the old resolver endpoint, making #1098 a **Minor Change** rather than a **Breaking Change**.

### Review Checklist

### DIFF
--- a/api/resolver.py
+++ b/api/resolver.py
@@ -43,11 +43,16 @@ from .dao.basecontainerstorage import ContainerStorage
 from .jobs import gears
 from .web.errors import APINotFoundException, InputValidationException
 
+def set_container_type(node, container_type):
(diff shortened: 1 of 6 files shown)


api/resolver.py
```diff
@@ -43,11 +43,16 @@
 from .jobs import gears
 from .web.errors import APINotFoundException, InputValidationException
 
+def set_container_type(node, container_type):
```
kofalt: Add a quick bit saying "this is here because old CLI", etc :)

PR #1289, opened 2018-07-12
Add pagination to DataViews

ehlertjd:
Specifically, support filtering, skip and limit. No support for sorting at this time, since data views are streamed.

Filtering is done in python as one of the last stages in the pipeline so that filtering on file columns can be supported. A possible optimization but complexity would be to do as much filtering during the aggregation stage as is possible.

Type coercion is done in the Filter stage to allow for meaningful comparison.

### Test Requirements
- Verify that pagination features work as expected
  - Filtering on various columns, using various comparison operators (covered by integration tests)
  - Skip and Limit (covered by integration tests)

### Review Checklist

### DIFF
--- a/api/data_views/config.py
+++ b/api/data_views/config.py
@@ -82,24 +82,11 @@ class DataViewConfig(object):
 
         max_idx = -1 
         for col in columns:
-            # Lookup src alias
-            src, datatype = ColumnAliases.get_column_alias(col['src'])
-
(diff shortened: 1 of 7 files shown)


ehlertjd:
Support `$in` concept comparing string to array

nagem:
Overall this looked good to me, would be interested in @ambrussimon's take when he has a chance to review.

PR #1296, opened 2018-07-16
No root 

hkethi002:
### Changes
Site admin can do everything root mode does
  - slight exception for GET list (root=true or exhaustive=true needed to return entire list regardless of permissions)

### Test Plan
Integration tests
  - Tested that as_admin and as_root do the same thing (When calling get_all used the exhaustive flag)
  - Lowered the testing level to user with admin access on the group/project in order to test the auth layer (admin now skips it like root)

Manual tests
  - Can see all projects with root=true (can also use exhaustive=true)
  - Api calls returns single project without root=true (Frontend blocks it out)
  - Can upload to projects without root-true or permissions as long as user is site admin

This is not a breaking change to the frontend
### Review Checklist

### DIFF
--- a/sdk/codegen/src/main/resources/fw-python/flywheel.mustache
+++ b/sdk/codegen/src/main/resources/fw-python/flywheel.mustache
@@ -59,8 +59,9 @@ class Flywheel:
         config = config_from_api_key(api_key)
         self.api_client = ApiClient(config, context=self)
 
-        # Root mode
+        # Root mode (Deprecated)
         if root:
+            log.warning('Root mode is deprecated')
             self.api_client.set_default_query_param('root', 'true')
 
         self.api_client.user_agent = 'Flywheel SDK/{} (Python {}; {})'.format(SDK_VERSION,
--- a/sdk/codegen/src/main/resources/fw-python/flywheel_methods.mustache
+++ b/sdk/codegen/src/main/resources/fw-python/flywheel_methods.mustache
@@ -52,7 +52,7 @@
         """
         return self.get_container(id, **kwargs)
 
-    def resolve(self, path):
+    def resolve(self, path, **kwargs):
         """Perform a path based lookup of nodes in the Flywheel hierarchy.
 
         :param str path: (required) The path to resolve
@@ -61,7 +61,7 @@
         if not isinstance(path, list):
             path = path.split('/')
 
-        return self.resolve_path(flywheel.ResolverInput(path=path))
+        return self.resolve_path(flywheel.ResolverInput(path=path), **kwargs)
 
     def lookup(self, path):
         """Perform a path based lookup of a single node in the Flywheel hierarchy.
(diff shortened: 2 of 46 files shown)


sdk/codegen/src/main/resources/fw-python/flywheel.mustache
```diff
@@ -45,18 +45,14 @@ def config_from_api_key(api_key):
     return config
 
 class Flywheel:
-    def __init__(self, api_key, root=False, skip_version_check=False):
+    def __init__(self, api_key, skip_version_check=False):
```
ehlertjd: Would keep `root` in the parameters, and print a deprecation warning if not present. 
Also, I think there should be a corresponding Matlab change.

sdk/codegen/src/main/resources/matlab/flywheel_methods.mustache
```diff
@@ -17,8 +17,8 @@
             [returnData, resp] = obj.modifyJob(jobId, struct('state', state), varargin{:});
         end
 
-        function [returnData, resp] = resolve(obj, path, varargin)
-            [returnData, resp] = obj.resolvePath(flywheel.model.ResolverInput('path', split(path, '/')), varargin{:});
+        function [returnData, resp] = resolve(obj, path, exhaustive, varargin)
```
ehlertjd: Would prefer to keep this as an optional, named parameter, which it *should* be if you just leave it as-is.

sdk/codegen/src/main/resources/fw-python/flywheel_methods.mustache
```diff
@@ -41,7 +41,7 @@
         """
         return self.modify_job(job_id, { 'state': state })
 
-    def resolve(self, path):
+    def resolve(self, path, exhaustive=False):
```
ehlertjd: I would just capture and pass `**kwargs` rather than explicitly adding the parameter.

sdk/codegen/src/main/resources/fw-python/flywheel.mustache
hkethi002: Do you mean if the root parameter is present?

sdk/codegen/src/main/resources/matlab/flywheel.mustache
```diff
@@ -47,6 +47,7 @@ classdef Flywheel < handle
 
             % Set root mode
             if exist('root', 'var') && root
+                fprintf('Root mode is deprecated')
```
ehlertjd: `fprintf('WARNING: Root mode is deprecated');`

sdk/codegen/src/main/resources/matlab/flywheel.mustache
```diff
@@ -47,7 +47,7 @@ classdef Flywheel < handle
 
             % Set root mode
             if exist('root', 'var') && root
-                fprintf('Root mode is deprecated')
+                fprintf('WARNING: Root mode is deprecated')
```
ehlertjd: Oh man I missed the `\n` at the end... Sorry!
`fprintf('WARNING: Root mode is deprecated\n');`

ehlertjd:
Tested with Matlab toolbox, and saw the warning and that query parameter was being forwarded. LGTM

ehlertjd:
Gave this another look, still LGTM

PR #1299, opened 2018-07-17
Support filtering data-view files on gear name and version

ehlertjd:
Utilizes the new `gear_info` subdoc on analyses, introduced in #1271.

### Review Checklist

### DIFF
--- a/api/data_views/data_view.py
+++ b/api/data_views/data_view.py
@@ -31,6 +31,12 @@ from .pipeline.missing_data_strategies import get_missing_data_strategy
 # TODO: subjects belong here once formalized
 SEARCH_CONTAINERS = ['projects', 'sessions', 'acquisitions']
 
+ANALYSIS_FILTER_COLUMNS = [
+    ('label', 'label'),
+    ('gear.name', 'gear_info.name'),
+    ('gear.version', 'gear_info.version')
+]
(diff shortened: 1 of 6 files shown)


api/data_views/data_view.py
```diff
@@ -135,12 +135,20 @@ def build_pipeline(self, output_format, pagination):
...
-                label_filter = config.analysis_filter.get('label')
-                match_analyses = MatchContainers('analysis', 'label', 'analysis', label_filter, match_type)
+                analysis_filters = []
+
+                if 'label' in config.analysis_filter:
+                    analysis_filters.append(('label', config.analysis_filter['label']))
+                if 'gear_name' in config.analysis_filter:
+                    analysis_filters.append(('gear_info.name', config.analysis_filter['gear_name']))
+                if 'gear_version' in config.analysis_filter:
+                    analysis_filters.append(('gear_info.version', config.analysis_filter['gear_version']))
```
nagem: Is there a reason these need to be a whitelist? Just so the user doesn't have to know about `gear_info`?

ehlertjd: Primarily, yes, it simplifies the definition a little bit. Although this block could be simplified to a list that's defined as a constant.

PR #1303, opened 2018-07-20
Save data view results to container

ehlertjd:
This PR introduces `/api/views/save` endpoint that allows executing a saved or ad-hoc data view, saving the result as a file (e.g. csv) on a container.  This uses SSE to report progress while the view data is being generated. Currently supports project, session and acquisition -- it may be interesting to add collections and analyses as destination containers.

The `FileCreator` class was added as an abstraction to write a file to a container using `TargetedMultiPlacer`. Please let me know if there was a better way that I missed.

### Test Requirements
Using SDK:
- Test saving data view results to various containers, in various formats
- Verify permissions, for the view being executed, the container being examined, and the container being updated.
- Verify access log contents for containers being accessed

### Review Checklist

### DIFF
--- a/api/create_file.py
+++ b/api/create_file.py
@@ -1,57 +1,93 @@
+"""Provides functionality for creating files on a container"""
 import datetime
 import uuid
 
 from . import config, files, util, upload
-from .placer import TargetedPlacer
+from .placer import TargetedMultiPlacer
+
+class PendingFile(object):
+    def __init__(self, filename, path, fileobj, metadata):
+        self.filename = filename
+        self.path = path
+        self.fileobj = fileobj
+        
+        if metadata is not None:
+            self.metadata = metadata.copy()
+        else:
+            self.metadata = {}
+        self.metadata['name'] = filename
 
 class FileCreator(object):
-    """Handle creation of a single file on target container"""
-    def __init__(self, handler):
+    """Helper class for creating file(s) on target container"""
+    def __init__(self, handler, container_type, container):
+        """Initialize the file creator, setting the destination container"""
         self.handler = handler
         self.file_processor = None
-        self.filepath = None
-        self.fileobj = None
-        self.filename = None
 
-    def create_file(self, filename):
-        self.filename = filename
+        self.container_type = container_type
+        self.container = container
+
+        self._files = []
+
+    def create_file(self, filename, metadata=None):
+        """Open a file for writing, and add it to the pending file list.
+
+        Arguments:
+            filename (str): The name of the file to create
+            metadata (dict): The optional metadata to set
+        """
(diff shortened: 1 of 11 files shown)


tests/integration_tests/python/test_data_views.py
```diff
@@ -1265,3 +1265,101 @@ def test_data_view_skip_and_limit(data_builder, file_form, as_admin):
...
+            { 'src': 'subject.code', 'dst': 'subject_label' },
+            { 'src': 'subject.age' },
+            { 'src': 'subject.sex' },
+            { 'src': 'session.label', 'dst': 'session_label' },
+            { 'src': 'acquisition.label', 'dst': 'acquisition_label' }
+        ]
+    }
+
+    # No public access
+    r = as_public.post('/views/save?containerId={}'.format(project), json={
```
nagem: I don't see the query param `containerId` being used in `execute_and_save`. Am I missing it? Can it use the value in the payload?

api/data_views/handlers.py
```diff
@@ -13,6 +15,9 @@
 
 log = config.log
 
+# TODO: Update when subjects are real
+FILE_CONTAINERS = { 'project', 'session', 'acquisition' }
```
nagem: FYI @ambrussimon

api/create_file.py
```diff
@@ -0,0 +1,92 @@
...
+
+        # Construct the file metadata list
+        metadata = [ pending_file.metadata for pending_file in self._files ]
+
+        # Create our targeted placer
+        placer = TargetedMultiPlacer(self.container_type, self.container, self.container['_id'],
+            metadata, timestamp, origin, {'uid': self.handler.uid}, self.file_processor, self.handler.log_user_access)
+
+        for pending_file in self._files:
+            # Not a great practice. See process_upload() for details.
```
nagem: Maybe eventually we can pull commonalities between `process_upload` and here into a func in `files.py`?

ehlertjd: Yeah, I think further refactoring is worthwhile, I wanted to limit the scope for this PR, knowing that files collection is coming.

tests/integration_tests/python/test_data_views.py
ehlertjd: This may be a little obtuse, but the query `containerId` query parameter is read in `do_execute_view` here:
https://github.com/flywheel-io/core/pull/1303/files/b12a8fcb23f6c538cda4437e0681650bd352c623#diff-5818f26e5755e7509dfa1b5a2dcf003bR169

We use distinct ids in the payload vs query parameter to permit you to execute a dataview against one container and save it to a different one. I'm open to revisiting whether or not that workflow makes sense.

PR #1305, opened 2018-07-23
Fix upgrade step to add gear_info to analyses

ehlertjd:
Should not be restricted by code-freeze since it fixes an issue in a merged PR.

This is a fix for DB upgrade to `52` where analyses were not getting `gear_info` updates because they carry the job id as a string rather than `ObjectId`.

To re-run the upgrade (needs to be done on dev.flywheel.io and next.dev.flywheel.io):
```
db.singletons.update({_id:'version'}, {$set: {database: NumberInt(51)}})
```

### Test Requirements

1. Reset db version to `51` as instructed above, then perform `reset-containers`
2. Verify that any analysis object that has an associated job has a `gear_info` subdocument.

Testing can be done via mongo client or SDK.

### Review Checklist

### DIFF
--- a/bin/database.py
+++ b/bin/database.py
@@ -1746,7 +1746,7 @@ def upgrade_job_to_52(job, gears):
     }}
     config.db.jobs.update_one({'_id': job_id}, update_doc)
     update_doc['$set']['gear_info']['id'] = gear_id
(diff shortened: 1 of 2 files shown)


nagem:
Let this PR be a reminder to us (me) that we shouldn't do piecemeal db policy changes (replacing _ids and FKs with strings).

PR #1312, opened 2018-07-25
tests batch with jobs, swagger fix

hkethi002:
Added batch jobs tests, found typo in swagger

### Review Checklist


ehlertjd:
Would it be a ton of extra effort to add the following as integration tests? (if they don't already exist in the core integration tests)

- Test adding jobs that aren’t valid
- Test adding jobs that will fail
- Test adding no jobs

hkethi002:
All of those cases are tested in the core integration batch tests

PR #1315, opened 2018-07-26
Add exclusive tags

kofalt:
This adds exclusive tags with a `!` prefix. To wit:

```
engine run -e -t '!dicom-mr-classifier,!dcm2niix,!qa-report-fmri'
```

### Test plan

Assume default job loadout.

Consume jobs with tags:

```
!dicom-mr-classifier,!dcm2niix,!qa-report-fmri
```

Observe no jobs consumed:

```json
"states": {
    "cancelled": 0,
    "complete": 0,
    "failed": 0,
    "pending": 37,
    "running": 0
},

```

Consume jobs with first tag of set removed:

```
!dcm2niix,!qa-report-fmri
```

Observe only classifier jobs consumed:

```json
"states": {
    "cancelled": 0,
    "complete": 37,
    "failed": 0,
    "pending": 37,
    "running": 0
},
```

Consume jobs with last tag inverted:

```
qa-report-fmri
```

Observe no change.

### DIFF
--- a/api/jobs/handlers.py
+++ b/api/jobs/handlers.py
@@ -395,6 +395,8 @@ class JobsHandler(base.RequestHandler):
     @require_admin
     def pending(self):
         tags = self.request.GET.getall('tags')
+
+        # Allow for tags to be specified multiple times, or just comma-deliminated
         if len(tags) == 1:
             tags = tags[0].split(',')
 
@@ -404,8 +406,10 @@ class JobsHandler(base.RequestHandler):
     def next(self):
         peek = self.is_true('peek')
         tags = self.request.GET.getall('tags')
-        if len(tags) <= 0:
-            tags = None
+
+        # Allow for tags to be specified multiple times, or just comma-deliminated
+        if len(tags) == 1:
+            tags = tags[0].split(',')
 
         job = Queue.start_job(tags=tags, peek=peek)
 
--- a/api/jobs/queue.py
+++ b/api/jobs/queue.py
@@ -309,9 +309,18 @@ class Queue(object):
         Potential jobs must match at least one tag, if provided.
         """
 
+        if tags is None:
+            tags = []
+
+        inclusive_tags = filter(lambda x: not x.startswith('!'), tags)
+        exclusive_tags =  map(lambda x: x[1:], filter(lambda x: x.startswith('!'), tags)) # strip the '!' prefix
+
         query = { 'state': 'pending' }
-        if tags is not None:
-            query['tags'] = {'$in': tags }
+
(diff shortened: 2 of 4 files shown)


tests/integration_tests/python/test_jobs.py
```diff
@@ -161,6 +161,10 @@ def test_jobs(data_builder, default_payload, as_public, as_user, as_admin, as_ro
     r = as_root.get('/jobs/next', params={'tags': 'fake-tag'})
```
ehlertjd: Add a test for `/jobs/next` with no tags.

api/jobs/handlers.py
```diff
@@ -404,8 +406,10 @@ def pending(self):
     def next(self):
         peek = self.is_true('peek')
         tags = self.request.GET.getall('tags')
-        if len(tags) <= 0:
-            tags = None
+
+        # Allow for tags to be specified multiple times, or just comma-deliminated
+        if len(tags) == 1:
```
ehlertjd: `if tags and len(tags) == 1:`

(Cannot take len of None)

kofalt: `getall` returns a list, this endpoint works with no tags

tests/integration_tests/python/test_jobs.py
kofalt: That is already covered, line 719 et al

api/jobs/queue.py
```diff
@@ -309,9 +309,18 @@ def start_job(tags=None, peek=False):
...
+        inclusive_tags = filter(lambda x: not x.startswith('!'), tags)
+        exclusive_tags =  map(lambda x: x[1:], filter(lambda x: x.startswith('!'), tags)) # strip the '!' prefix
+
         query = { 'state': 'pending' }
-        if tags is not None:
-            query['tags'] = {'$in': tags }
+
+        if len(inclusive_tags) > 0:
+            query['tags'] = {'$in': inclusive_tags }
+        if len(exclusive_tags) > 0:
```
nagem: Is the expected outcome that only the exclusive tags are used if a user supplies both? I'd rather we `400` if the both lists are non-empty.

kofalt: Is there a reason you wouldn't want both?

kofalt: Whoop, I see what you mean, this over-assigns. Changing!

kofalt:
Good catch by Megan: updated so that inclusive & exclusive tags can coexist.

kofalt:
Travis threw its toys out of the pram; rebasing for my own amusement.

PR #1329, opened 2018-07-31
Group post

### LINKED ISSUE #1304: POSTing an existing group sets created and modified time to current time
When a group is being created, update_one is called with upsert=true. This causes the created and modified timestamps to be bumped when a group is created when a group with the same id already exists.

hkethi002:
Fixes #1304 
### Review Checklist


nagem:
/shrug. It's not breaking (still allows POST as upsert replace), but fixes the issue with `created` time. We should still resolve if an upsert POST is needed.

hkethi002:
The folder reaper at least seems to expect to be able to upsert groups with POST

PR #1334, opened 2018-08-03
Re-use fixtures between unit and integration tests

davidfarkas:
With this the unit and integration test fixtures are importable with pytest like `pytest_plugins = ['tests.fixtures.unit_tests', 'tests.fixtures.common']`. There is also a set of common fixtures `tests.fixtures.common` which are re-usable between the unit and integration tests.
It was necessary to add integration tests to `coreplus` without duplicating the fixtures again.

### Review Checklist


nagem:
Same sentiment as @ehlertjd. I scanned the functions I'm more familiar with and didn't notice any changes.

PR #1341, opened 2018-08-07
Rerun rule-spawned jobs on replaced files

### LINKED ISSUE #878: Possibly update created date, replace all file fields on file replace
_Friday Jul 28, 2017 at 16:14 GMT_
_Originally opened as https://github.com/scitran/core/issues/878_

----

When a file with the same name as an existing file in a container is uploaded, it replaces the file itself and updates the file attributes stored in Mongo, preserving some of the old values like `created`.  It is possible we'd want to replace all fields and treat it as a new file. Further thought/discussion required.

### LINKED ISSUE #1081: Replacement of a file should trigger gear rule
<a href="https://github.com/ryansanford"></img></a> **Issue by [ryansanford](https://github.com/ryansanford)**
_Monday Feb 19, 2018 at 19:05 GMT_
_Originally opened as https://github.com/scitran/core/issues/1081_

----

A file was replaced with updated contents via the upload/reaper endpoint, resulting in the measurements and info fields being cleared. This replacement happened after the dicom-mr-classifier gear completed. The result was that the info and measurements fields were left empty, and the gear rule in place did not fire a new job to reconstitute their contents. Also, the generated nifti files are not in sync with the latest dicom.

nagem:
Closes #1081 
Closes #878

_Copied from `test_rules_rerun_after_file_replace` function sig:_

Always run jobs from rules where at least one of the inputs were "replaced" during the upload.

When a file is uploaded to a container that already has a file with the same name, the original file is completely replaced with a new one in mongo. 
      * Except when the upload type is "reaper", and the hash is the same. In that case the file upload is ignored.
      ** See comment in code for discussion about created/replaced timestamp.

In the rule evaluation step, the rules are evaluated against the container's state before the upload (set A) and after (set B). Jobs in set B that are not in set A are queued if no files were replaced during the upload.

If files were replaced, any job in set A that has a replaced file as input is also queued.

### Breaking Changes
None, although rules will spawn jobs when they had not before

### Affected Components
File Upload, Reaper Upload, Rules

### Test Plan Overview 
  - Upload files via the different placer endpoints, ensure entire file object is replaced, created date stays the same, replaced date is added
  - Upload files via `/upload/reaper`. Make sure replaces of the same file with the same hash are ignored.
  - Ensure expected rules trigger as if the file was uploaded for the first time
  - Test reaper uploads on ObjS
  
### Checklist
- [x] Integration/Unit tests
- [x] Swagger updates and changes (just for replaced field if it sticks around)
- [x] Test Plan Documentation

### CCs
@ryansanford @gsfr @tcbtcb


api/dao/basecontainerstorage.py
```diff
@@ -220,7 +220,10 @@ def get_parent(self, _id, cont=None, projection=None):
 
 
     def _from_mongo(self, cont):
-        pass
+        # For now, don't return information about a file's previous version(s)
+        if cont:
+            for f in cont.get('files', []):
+                f.pop('other_versions', None)
```
nagem: The removal of this must not have made it into the commit, fixing....

api/dao/hierarchy.py
```diff
@@ -231,69 +193,125 @@ def check_cont(cont, reqs):
...
-            log.warn('Fileinfo passed with non-integer size')
-            fileinfo['size'] = int(fileinfo['size'])
+
+def replace_file(cont_name, _id, existing_fileinfo, fileinfo, access_logger):
+
+    access_logger(AccessType.replace_file, cont_name=cont_name, cont_id=_id, filename=fileinfo['name'])
+
+    # Keep created date the same, add "replaced" timestamp
+    fileinfo['created'] = existing_fileinfo['created']
+    fileinfo['replaced'] = datetime.datetime.utcnow()
```
nagem: Help Wanted.

I decided to keep the original created date and add a `replaced` timestamp. Open to feedback about this approach.

tests/integration_tests/python/test_rules.py
```diff
@@ -743,3 +743,278 @@ def test_auto_update_rules(data_builder, api_db, as_admin):
         'gear_id': gearv1
     })
     assert r.status_code == 400
+
```
nagem: This test is really long and a bit extra but I wanted to test the rerun of the full `file lands -> classify -> convert -> qa` that is the default for most of our users. I think it's beneficial to have something like this as a workflow in the integration tests regardless of if it is necessarily needed to test the rerun logic.

api/dao/hierarchy.py
hkethi002: I think that is a good approach, but I think the frontend should reflect that, right now it only provides the `Uploaded` date which is the `created` timestamp, which is a bit confusing now.

api/dao/hierarchy.py
```diff
@@ -231,69 +193,125 @@ def check_cont(cont, reqs):
...
+                container_after = add_file(cont_name, _id, fileinfo)
+                saved_state = 'saved'
+
+
+            # Files from a failed job should never replaced existing files that are "accepted" (unless they are deleted)
+            elif fileinfo.get('from_failed_job') and not f.get('from_failed_job'):
+                saved_state = 'ignored'
+
+            # The file object is the same as an existing file and the caller has chosen to ignore in this situation
+            elif f.get('hash') == fileinfo['hash'] and ignore_hash_replace:
```
ehlertjd: I wonder if we should treat a non-existent hash on either side as if the hashes *didn't* match.

nagem: A non-existent hash on one side but not the other would be a non-match. But if _both_ evaluate to `None` it would be like they were identical, which is probably not ideal. Not sure the likelihood of us missing a hash from the new file's attributes.

api/dao/hierarchy.py
```diff
@@ -22,50 +22,12 @@
...
+        self.dbc = config.db[level]
+        self.id_ = container['_id']
+        self.file_prefix = 'files'
 
-    def add_file(self, fileinfo):
-        return self.dbc.find_one_and_update(
-            {'_id': self.id_},
-            {'$push': {self.file_prefix: fileinfo}},
-            return_document=pymongo.collection.ReturnDocument.AFTER
-        )
```
nagem: I removed dead code here FYI.

api/jobs/rules.py
```diff
@@ -247,13 +252,25 @@ def create_jobs(db, container_before, container_after, container_type):
     # Using a uniqueness constraint, create a list of the set difference of jobs_after \ jobs_before
     # (members of jobs_after that are not in jobs_before)
     for ja in jobs_after:
-        new_job = True
-        for jb in jobs_before:
-            if ja['job'].intention_equals(jb['job']):
-                new_job = False
-                break # this job matched in both before and after, ignore
-        if new_job:
+
```
nagem: Below is where the `set().intersect(set())` logic lived that I replaced by a for loop with an `in` without issue.

api/placer.py
```diff
@@ -202,21 +234,10 @@ def process_file_field(self, field, file_attrs):
         r_metadata  = target['metadata']
         file_attrs.update(r_metadata)
 
-        if container.level != 'subject':
```
nagem: Removed dead subject file code.

tests/unit_tests/python/test_uploads.py
```diff
@@ -42,7 +42,7 @@ def test_signed_url_reaper_upload(as_drone, mocker):
     ticket_id = r.json['ticket']
```
nagem: These tests are failing and I'm not sure why. Commented out for now, will reevaluate with @ambrussimon.

api/dao/hierarchy.py
ehlertjd: I guess I would check whether or not reapers using signed urls to upload are generating a hash. I was under the impression that they weren't, but I could be wrong.

api/dao/hierarchy.py
ehlertjd: Update json schema for the 'replaced' attribute

api/upload.py
```diff
@@ -144,9 +144,9 @@ def process_upload(request, strategy, access_logger, container_type=None, id_=No
         # Stands in for a dedicated object... for now.
         file_attrs = {
             '_id': field.uuid,
-            'name': field.filename,
-            'modified': field.modified,
-            'size': field.size,
+            'name':	 field.filename,
+            'modified': timestamp,
```
kofalt: why this change?

nagem: I think that is leftover from when this was changed more significantly for version support. Line 141 sets it to the same thing so no functional change. I'll set it back.

PR #1343, opened 2018-08-08
Subjects collection

### LINKED ISSUE #389: Add explicit subject entity
<a href="https://github.com/gsfr"></img></a> **Issue by [gsfr](https://github.com/gsfr)**
_Saturday Aug 06, 2016 at 19:20 GMT_
_Originally opened as https://github.com/scitran/core/issues/389_

----

DO NOT WORK ON THIS! DO NOT CLOSE! Capturing discussion outcome.

Adding a distinct subject entity to our data model and endpoint may be required to achieve certain subject related functionality, including proper BIDS support.

Pros:
- full BIDS support
- subject-level files and metadata
- subject centric view

Cons:
- more hierarchy
- more joins

Data Model

**Subject**
- _id
- project (project pointer)
- date and place of birth
- files []
- info {}

**Session**
- _id
- project (project pointer)
- subject (subject pointer, may be null)
- transient (today's subject info)
  - first name, last name 
  - age, weight, sex, etc

Not sure if race and ethnicity are permanent or transient subject attributes.

cc @nagem, @rentzso

ambrussimon:
Resolves #389 

Subjects just moved to their own mongo collection.
Sorry about the hefty diff, couldn't really bite it in smaller commits.

* pro: lot of quirky code was stripped that handled embedded subjects differently to standard containers
* con: lot of quirky code was added that enables backwards-compatibility (ie. CRUD via sessions)
* still need to fix SDK tests and enable subject jobs

### Review Checklist


swagger/schemas/mongo/subject.json
```diff
@@ -6,8 +6,14 @@
...
 
-    "age":              { "type": ["number", "null"] },
+    "project":          {},
+    "created":          {},
+    "modified":         {},
+    "permissions":      {},
+
     "sex":              { "enum": ["male", "female", "other", "unknown", null] },
+    "type":             { "enum": ["human", "animal", "phantom", null] },
+    "species":          { "type": "string" },
```
ehlertjd: Missing `cohort`?

AshleyAncona:
@ambrussimon @nagem we need to ensure that the project report still works properly after all these changes.

swagger/schemas/definitions/subject.json
```diff
@@ -66,6 +67,7 @@
             "type": "object",
             "properties": {
               "_id":              {"$ref":"#/definitions/_id"},
+              "project":          {"$ref":"#/definitions/_id"},
```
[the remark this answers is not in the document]
ambrussimon: Right, thanks. Analysis/job support added in latest commit.

swagger/schemas/mongo/subject.json
ambrussimon: I don't remember `cohort` being mentioned originally for this scope; but am happy to add, just please verify that's desired here.

api/config.py
```diff
@@ -177,6 +177,7 @@ def initialize_db():
     db.sessions.create_index([('project', 1), ('label', 1)])
```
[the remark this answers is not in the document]
ambrussimon: Thanks, fixed.

swagger/schemas/mongo/subject.json
ehlertjd: Yeah, it's in the feature card:
> * new fields:
> ...
>   * cohort
>     * control
>     * subject

api/dao/basecontainerstorage.py
```diff
@@ -471,6 +473,20 @@ def join_origins(container):
...
                     container['join-origin'][j_type][j_id] = join_doc
 
+    @staticmethod
+    def join_subjects(sessions):
+        """Given a list of sessions, join their subjects."""
+        storage = ContainerStorage.factory('subjects')
+        query = {'_id': {'$in': list(set(sess['subject'] for sess in sessions))}}
+        subjects = {subj['_id']: subj for subj in storage.get_all_el(query, None, None)}
+        for session in sessions:
+            subject = subjects[session['subject']]
```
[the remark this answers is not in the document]
ambrussimon: Yes, the upgrade script (similarly to previous subject-related upgrade 44) groups sessions together that have no/empty subject, no/empty subject.code and makes sure they are assigned different ids.

Beyond that, my understanding is that currently all ways a session can get added/modified, it will have a subject eventually.

ambrussimon:
@ehlertjd 
> 1. Is this the right PR to do swagger documentation for subjects?

I assume you mean `/subjects/*`, and definitely. Will do RN.

> 2. Is this the right PR to update the resolver?

I'm not sure, that could be considered breaking. I'll implement it with a feature-flag, then.

> 3. Do the generic container endpoints work with subjects? Can we add tests to verify that?

Yes, and yes. (Already added, though not exhaustively.)

@AshleyAncona 
Project reports should be functional, let me know if something is of. Tests should definitely be extended in that case.

ambrussimon:
This is ready for review - now there's nothing missing AFAIK.

sdk/src/matlab/testDrive.m
```diff
@@ -99,7 +137,7 @@
 sessionDownloadFile = fullfile(tempdir, 'download2.txt');
 fw.downloadFileFromSession(sessionId, filename, sessionDownloadFile);
 
-session = fw.getSession(sessionId);
+session = fw.getSesSubjectsion(sessionId);
```
nagem: Looks like there was a typo here

bin/database.py
```diff
@@ -1823,6 +1824,98 @@ def upgrade_to_54():
...
+            'parents': session['parents'],
+            'project': session['project'],
+            'permissions': session['permissions']
+        })
+        session['subject'] = subject['_id']
+        if subject.get('race') or subject.get('ethnicity'):
+            subject['type'] = 'human'
+        if subject.get('age'):
+            session['age'] = subject.pop('age')
+        return subject
```
nagem: Subjects are missing `created` and `modified` after creation.

swagger/schemas/definitions/subject.json
```diff
@@ -43,13 +55,19 @@
...
 
               "age":              {"$ref":"#/definitions/age"},
+
               "sex":              {"$ref":"#/definitions/sex"},
+              "cohort":           {"$ref":"#/definitions/cohort"},
+              "type":             {"$ref":"#/definitions/type"},
               "race":             {"$ref":"#/definitions/race"},
               "ethnicity":        {"$ref":"#/definitions/ethnicity"},
+              "species":          {"type": ["string", "null"]},
+              "strain":           {"type": ["string", "null"]},
```
nagem: We should add `label` to the input schema

api/download.py
```diff
@@ -145,32 +145,27 @@ def _preflight_archivestream(self, req_spec, collection=None):
 
                 sessions = config.db.sessions.find({'project': item_id, 'deleted': {'$exists': False}}, ['label', 'files', 'uid', 'timestamp', 'timezone', 'subject'])
                 session_dict = {session['_id']: session for session in sessions}
-                acquisitions = config.db.acquisitions.find({'session': {'$in': session_dict.keys()}, 'deleted': {'$exists': False}}, ['label', 'files', 'session', 'uid', 'timestamp', 'timezone'])
                 session_prefixes = {}
 
+                subjects = config.db.subjects.find({'_id': {'$in': list(set(sess['subject'] for sess in session_dict.itervalues()))}})
```
hkethi002: Since we're anyways downloading subjects the same way as other containers when we download, we should allow subject nodes to be passed as well.

nagem:
I noticed subject files aren't being accounted for in the "summarize" step of downloads: https://github.com/flywheel-io/core/pull/1343/files#diff-5f6b428f8ab6dc5134ff7b9cbcad2062R376

nagem:
So you don't have to scan the PR looking for comments I had left and whether or not you had addressed them, I'm going to compile them in a list here:
  - [ ] Add "label" to input schema for subjects
  - [ ] `created` and `modified` are not populated on the subject during db update 
  - [ ] The `summary` step in `download.py` does not factor in subject files
  - [ ] Update the `cleanup_deleted.py` script

hkethi002:
I'll copy Megan as well:
- [ ] Add `/projects/{ProjectId}/subjects` to projects.yml in swagger
- [ ] Allow subjects to be a download node

nagem:
Final thought and then I believe my review is complete: 

Do we want to store `label` on the subject and populate `code` artificially instead? My main motivator for wanting to switch to `label` is to avoid situations like [this](https://github.com/flywheel-io/core/pull/1343/files#diff-33d8c0d881a2cdd1e055210032175cf0R45). Doesn't necessarily need to be in this PR.

bin/database.py
ambrussimon: They are populated a bit later; see [L1903](https://github.com/flywheel-io/core/pull/1343/files#diff-837f544139dd99be7b888e157752643bR1903). Added tests, though.

ambrussimon:
* even though `subject.label` is swagger'd, it's not too useful w/o api support
* code/label overhaul was greatly desired throughout, but indeed didn't fit scope

api/handlers/containerhandler.py
```diff
@@ -357,33 +357,25 @@ def put(self, cont_name, **kwargs):
...
-
-            # Enable subject matching on session updates
-            if not payload.get('subject', {}).get('_id') and not payload.get('subject', {}).get('code'):
-                match_key = 'code' if container['subject'].get('code') else '_id'
-                payload.setdefault('subject', {})[match_key] = container['subject'][match_key]
+            subject_code = target_subject_code or container['subject'].get('code')
+
+            # Check for subject code collision 1st when updating project and/or subject code
+            if subject_code and (target_parent_container or target_subject_code):
+                if config.db.subjects.find_one({'project': parent_container['_id'], 'code': subject_code}):
```
nagem: Can you use `get_all_el` or add the deleted filter here?

api/handlers/containerhandler.py
```diff
@@ -334,22 +349,60 @@ def put(self, cont_name, **kwargs):
...
+                target_subject = config.db.subjects.find_one({'_id': bson.ObjectId(target_subject_id)})
+                if not target_subject:
+                    raise APINotFoundException('cannot find subject {}'.format(target_subject_id))
+                if target_subject.get('deleted'):
+                    raise APINotFoundException('subject {} is marked for deletion'.format(target_subject_id))
+                if target_subject['project'] != parent_container['_id']:
+                    raise APINotFoundException('subject {} is not in project {}'.format(target_subject_id, parent_container['_id']))
+
+            # Handle changing project: copy subject into target project if there are other sessions on it (otherwise move)
+            elif target_parent_container and config.db.sessions.count({'subject': container['subject']['_id']}) > 1:
```
nagem: This action happens before permissions to do the update are checked.

nagem:
One more minor request from @gsfr:

The cohort list should be changed to: Control, Study (instead of Subject), Training, Test, Validation

api/handlers/containerhandler.py
```diff
@@ -334,22 +349,56 @@ def put(self, cont_name, **kwargs):
...
+            # Handle changing subject
+            if target_subject_id:
+                if not bson.ObjectId.is_valid(target_subject_id):
+                    raise APIValidationException('{} is not a valid object id'.format(target_subject_id))
+                target_subject = config.db.subjects.find_one({'_id': bson.ObjectId(target_subject_id)})
+                if not target_subject:
+                    raise APINotFoundException('cannot find subject {}'.format(target_subject_id))
+                if target_subject.get('deleted'):
+                    raise APINotFoundException('subject {} is marked for deletion'.format(target_subject_id))
+                if target_subject['project'] != parent_container['_id']:
```
nagem: Is there a reason we want to prevent this? Shouldn't we just update the project/parents list?

nagem:
@LinaJohnson and I found a few subjects with a string `project` key rather than an ObjectId. Haven't been able to reproduce that yet, and I didn't see anywhere in the code that would be coming from, so it might have been a previous bug. @ambrussimon could you take a look in the likely places you can think of? We'll update if we see it again.

api/handlers/containerhandler.py
```diff
@@ -334,22 +352,65 @@ def put(self, cont_name, **kwargs):
...
+            if target_subject_id:
+                target_subject = subject_storage.get_container(target_subject_id)
+
+                # If payload also contains project, verify that the target_subject is in it
+                if target_project and project_id != target_subject['project']:
+                    raise APIValidationException('subject {} is not in project {}'.format(target_subject_id, project_id))
+
+                # Make sure session.project is also updated
+                payload['project'] = target_subject['project']
+
```
hkethi002: target_project should be set here to the target subjects project, otherwise the target_subject's project gets overwritten in the call to _handle_embedded_subject()

api/handlers/containerhandler.py
```diff
@@ -291,6 +304,11 @@ def post(self, cont_name):
         if payload.get('timestamp'):
             payload['timestamp'] = dateutil.parser.parse(payload['timestamp'])
         permchecker = self._get_permchecker(parent_container=parent_container)
+
+        # Handle embedded subjects for backwards-compatibility
+        if cont_name == 'sessions':
+            self._handle_embedded_subject(payload, parent_container)
```
hkethi002: The project id in the payload is not an ObjectId and I think that is what's causing the project id strings in some subjects

api/handlers/containerhandler.py
ambrussimon: Thanks, fixed.

api/handlers/containerhandler.py
ambrussimon: I jumped on this, but the `subject.project` is actually populated from `parent_container` in `extract_subject()`, so I think it's not the culprit. Trying to reproduce in a failing test with the workflow you mentioned.

ambrussimon:
@nagem @hkethi002 I fixed one of the bugs Harsha mentioned, and still couldn't reproduce the str subject.project. I added a sort of defensive ObjectId casting in `extract_subject()`, which has a good chance of solving it.
Also, could someone please look at `dataexplorerhandler.py`? I realized that wasn't touched yet, and don't have the expertise to fix it quickly. (I'm not even sure it _needs_ to be updated, but naively I'd assume so.)

hkethi002:
It shouldn't need any changes, I've made the necessary changes to mongo-connector, and so far testing it on demo-delete, it seems to be working with all the subject changes.

PR #1346, opened 2018-08-08
Only check for major version differences by default

ehlertjd:
The SDK version check is too noisy, especially when working against dev projects. For now, toggle everything except for the major version comparison.

Can be re-enabled by passing `skip_version_check=False` or setting the environment variable:
`FLYWHEEL_SDK_SKIP_VERSION_CHECK=0`

### Review Checklist

### DIFF
--- a/sdk/codegen/src/main/resources/fw-python/flywheel.mustache
+++ b/sdk/codegen/src/main/resources/fw-python/flywheel.mustache
@@ -66,7 +66,9 @@ class Flywheel:
         {{/apis}}
 
         # Perform version check
-        self.check_version = not skip_version_check or '0' == os.environ.get('FLYWHEEL_SDK_SKIP_VERSION_CHECK')
+        self.check_version = not skip_version_check
+        if not self.check_version:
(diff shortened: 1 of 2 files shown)


sdk/codegen/src/main/resources/fw-python/flywheel.mustache
```diff
@@ -66,7 +66,7 @@ class Flywheel:
         {{/apis}}
 
         # Perform version check
-        self.check_version = not skip_version_check and not os.environ.get('FLYWHEEL_SDK_SKIP_VERSION_CHECK')
+        self.check_version = not skip_version_check or '0' == os.environ.get('FLYWHEEL_SDK_SKIP_VERSION_CHECK')
```
kofalt: Ugh, I would expect to be able to set the env to false, suggest using the version that also returns if the env is set & checking for falseness.

kofalt: I would expect to be able to set the env to false, suggest using the version that also returns if the env is set & checking for falseness.

PR #1350, opened 2018-08-09
Filter out gears that are disabled

ehlertjd:
Simply filter gears from the list endpoint where `gear.custom.flywheel.disabled` is `true`.

Tested with integration tests, as well as hand-testing creating rules, run and batch run on local instance with dev DB backup.

### Review Checklist

### DIFF
--- a/api/jobs/gears.py
+++ b/api/jobs/gears.py
@@ -24,7 +24,10 @@ def get_gears(all_versions=False, pagination=None):
     """
 
     if all_versions:
-        kwargs = {'sort': [('gear.name', 1), ('created', -1)]}
+        kwargs = {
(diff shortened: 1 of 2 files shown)


kofalt:
I don't think this is sufficient - we should also block an added job from launching. It would be nice to get a little more time to review...

kofalt:
Further addendum, the key should be `invalid` not `disabled.
https://github.com/flywheel-io/gears/tree/master/spec#reserved-custom-keys

Already blocked from launching here:
https://github.com/flywheel-io/core/blob/2330e69af9afae4d9e18a5ddc31cdc4e1b962b8c/api/jobs/queue.py#L176-L177

PR #1352, opened 2018-08-10
Use 'invalid' keyword for disabled gears

ehlertjd:
See #1350 - Used incorrect keyword for gears that shouldn't be runnable.

### Review Checklist


kofalt:
Threw on a doc commit. Hope you don't mind.

ehlertjd:
Good call, thanks

PR #1353, opened 2018-08-10
Auth config is replace instead of update

### LINKED ISSUE #1338: SCITRAN_AUTH_CONFIG_FILE doesn't allow removal of auth scheme.
Removing google section in file specified by `SCITRAN_AUTH_CONFIG_FILE` doesn't cause that section to be removed from the config.

Expect to be able to manage the removal more often as we lean on google initially for installation and training, then disable google at customer request once site-specific authentication has been validated.

Open to a different way all-together for Ops to manage Auth config. The existing file-based mechanism is nice for its simplicity, and similar to how other components are configured.

hkethi002:
Fixes #1338 
### Review Checklist

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -80,12 +80,10 @@ def apply_env_variables(config):
 
     # Load auth config from file if available
     if 'SCITRAN_AUTH_CONFIG_FILE' in os.environ:
-        auth_config = config['auth']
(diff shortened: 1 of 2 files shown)


nagem:
We should make sure no site is expecting any section of the auth config to be provided for them if they do not supply it in an environment variable. I doubt it, since the google auth in the default config is not valid anyway.

PR #1354, opened 2018-08-13
Pluggable gear transport

kofalt:
1. Remove gear prefetch feature. It's a red-headed stepchild, and we've changed our tradeshow strategy to not use this anymore.
1. Allow the gear scheme to be non-http. Unlike http URIs, the scheme component is separated out into the input `type`.

### DIFF
--- a/api/jobs/jobs.py
+++ b/api/jobs/jobs.py
@@ -6,6 +6,7 @@ import bson
 import copy
 import datetime
 import string
+from urlparse import urlparse
 
 from ..types import Origin
 from ..dao.containerutil import create_filereference_from_dictionary, create_containerreference_from_dictionary, create_containerreference_from_filereference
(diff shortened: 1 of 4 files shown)


api/jobs/jobs.py
```diff
@@ -278,11 +279,22 @@ def generate_request(self, gear):
...
+        uri = gear['exchange']['rootfs-url']
+        parsed = urlparse(uri)
+        scheme = parsed.scheme
+
+        if scheme == 'https':
+            # SSL does not change the input scheme type, both are just 'http'
+            scheme = 'http'
+        else:
+            # Other URI types keep the input scheme separate
+            uri = parsed.netloc + parsed.path
```
ehlertjd: Precludes username/password authentication, or anything in the query parameters. May be worth documenting the limitation - don't ask me where.

kofalt: True. The module doesn't have a "reconstruct" function so I'm going with "bare minimum until it becomes a problem". We implicitly require authentication be provided separately, so

PR #1359, opened 2018-08-14
Batch Optional Inputs

hkethi002:
### Changes
New `optional_input_policy` is a required key for batch proposals if the gear has optional inputs
- `ignored`
  - Optional inputs are always ignored and will not contribute to whether a container is matched nor will it be used in the job as an input
  - Won't contribute to whether the container has ambiguous inputs
- `flexible`
  - If a file is in the container that is suggested for the optional input, it will be used, if not, the container will still be considered a match and that job will just run without the optional input
  - The container may be marked as ambiguous if multiple files are suggested for the optional input
- `required`
  - Optional inputs are treated the same as required inputs
  - This is current functionality

Fixes bug where gears that have file inputs and api-key inputs 500 on batch runs
### Required frontend changes
If a gear has optional inputs, the user must specify an `optional_input_policy`
### Review Checklist

### DIFF
--- a/api/jobs/batch.py
+++ b/api/jobs/batch.py
@@ -87,8 +87,11 @@ def find_matching_conts(gear, containers, container_type, optional_input_policy,
                 opt_ignore = optional_input_policy == 'ignored' and is_optional_input
                 opt_required = optional_input_policy == 'required' or not is_optional_input
 
+                # Skip ambiguity check for this input if the policy is to ignore and the input is optional
                 if len(files) > 1 and not opt_ignore:
                     ambiguous = True
+                # Skip the not_matched check for this input if the policy is to ignore or to be flexible
+                # and the input is an optional input
                 elif opt_required and len(files) == 0:
                     not_matched = True
                     break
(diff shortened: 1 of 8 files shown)


api/jobs/batch.py
```diff
@@ -79,12 +80,16 @@ def find_matching_conts(gear, containers, container_type, context_inputs=False,
             suggestions = gears.suggest_for_files(gear, files, context=context)
 
             # Determine if any of the inputs are ambiguous or not satisfied
-            ambiguous = False # Are any of the inputs ambiguous?
+            ambiguous = False  # Are any of the inputs ambiguous?
             not_matched = False
-            for files in suggestions.itervalues():
-                if len(files) > 1:
+            for input_name, files in suggestions.iteritems():
```
nagem: Mind just writing a bit here about how the optional input policy works?

nagem: Should probably be in the swagger as well. Doesn't need to be a huge writeup, just describing the three different options and what they do.

ehlertjd:
Does this *have* to be a breaking change? What's wrong with defaulting to `required` if not specified?

PR #1362, opened 2018-08-16
Custom gear UID/GID

kofalt:
This enables a small trial feature to launch as a custom UID/GID. We want to test this out with internal gears first, so it will not be documented / moved out of `custom` unless we like it.

@ehlertjd I'm not sure if the gear request object ends up in swagger definitions, mind helping out?

### DIFF
--- a/api/jobs/jobs.py
+++ b/api/jobs/jobs.py
@@ -323,7 +323,7 @@ class Job(object):
             r['target']['env'][key] = gear['gear']['environment'][key]
 
         # Add command, if any
-        command_base = 'env; rm -rf output; mkdir -p output; '
+        command_base = ''
         if gear['gear'].get('command') is not None:
 
             command = render_template(gear['gear']['command'], self.config['config'])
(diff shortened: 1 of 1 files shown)


api/jobs/jobs.py
```diff
@@ -323,7 +330,7 @@ def generate_request(self, gear):
             r['target']['env'][key] = gear['gear']['environment'][key]
 
         # Add command, if any
-        command_base = 'env; rm -rf output; mkdir -p output; '
+        command_base = ''
```
nagem: Why was this in here before and what changed to allow it's removal?

nagem: Actually, I did see this: https://github.com/flywheel-io/engine/pull/95/files#diff-f54077dea3d06228ed0f57b318417ad2R253

If this is a breaking change, does this mean the API is now only compatible with engine-2?

kofalt: Let's open the history books :) :gear: :book: :gear: 

The `env` was pure debugging noise that we shouldn't emit; gears are free to take the same action if they want.

The rest was an ancient, preemptive protection around "undefined behavior" involving output folder state, which in practice was not necessary. The gear builder has long since made sure that folder is not included in the upload, and the engine now enforces this state in a more authoritative way.

PR #1363, opened 2018-08-16
Sdk gear rules

hkethi002:
### Changes
- new key for projects, sessions, acquisitions, and analyses: `parents`
  - ```parents: {project: projectId, group: groupId, ...}```
  - Containers will only have the applicable keys in `parents`
- New type of job api key
  - stores takes a `scope: {level: cont_type, id: cont_id, access: permission level}`
- jobs with tag `auto` that have an api key input get the above type of key
  - scope is set to `{level: project, id: projectId, access: ro}`
- auth layer checks for handler.scope (which is set from the api key)

- [x] __Note to self__: Change tests to not use db updates to set read-only once gear spec is updated
- [x] Add tests for optional inputs in rules
### Review Checklist

### DIFF
--- a/api/auth/__init__.py
+++ b/api/auth/__init__.py
@@ -34,7 +34,7 @@ def _check_scope(scope, container, parent_container=None):
     if container and container.get('parents'):
         return scope['id'] in container['parents'].itervalues() or scope['id'] == container['_id']
     elif parent_container and parent_container.get('parents'):
-        return scope['id'] in parent_container['parents'].itervalues()
+        return scope['id'] in parent_container['parents'].itervalues() or scope['id'] == parent_container['_id']
 
 def has_access(uid, container, perm):
     return _get_access(uid, container) >= INTEGER_PERMISSIONS[perm]
--- a/api/auth/__init__.py
+++ b/api/auth/__init__.py
@@ -19,8 +19,11 @@ PERMISSIONS = [
 INTEGER_PERMISSIONS = {r['rid']: i for i, r in enumerate(PERMISSIONS)}
 
 def _get_access(uid, container, scope=None):
-    if scope and _check_scope(scope, container):
-        return INTEGER_PERMISSIONS[scope['access']]
+    if scope:
+        if _check_scope(scope, container):
+            return INTEGER_PERMISSIONS[scope['access']]
+        else:
+            return -1
     permissions_list = container.get('permissions', [])
     for perm in permissions_list:
         if perm['_id'] == uid:
@@ -28,15 +31,10 @@ def _get_access(uid, container, scope=None):
     return -1
 
 def _check_scope(scope, container, parent_container=None):
-    if scope:
-        # print "Scoped request"
-        if container and container.get('parents'):
-            # print type(scope['id'])
-            # print type(container['_id'])
-            return scope['id'] in container['parents'].itervalues() or scope['id'] == container['_id']
-        elif parent_container and parent_container.get('parents'):
-            return scope['id'] in parent_container['parents'].itervalues()
-    return True
+    if container and container.get('parents'):
+        return scope['id'] in container['parents'].itervalues() or scope['id'] == container['_id']
+    elif parent_container and parent_container.get('parents'):
+        return scope['id'] in parent_container['parents'].itervalues()
 
 def has_access(uid, container, perm):
     return _get_access(uid, container) >= INTEGER_PERMISSIONS[perm]
--- a/api/auth/__init__.py
+++ b/api/auth/__init__.py
@@ -18,12 +18,25 @@ PERMISSIONS = [
 
 INTEGER_PERMISSIONS = {r['rid']: i for i, r in enumerate(PERMISSIONS)}
 
-def _get_access(uid, container):
+def _get_access(uid, container, scope=None):
+    if scope and _check_scope(scope, container):
+        return INTEGER_PERMISSIONS[scope['access']]
     permissions_list = container.get('permissions', [])
     for perm in permissions_list:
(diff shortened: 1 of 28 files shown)


api/dao/hierarchy.py
```diff
@@ -39,6 +39,12 @@ def get_container(cont_name, _id):
         '_id': _id,
     })
 
+def get_parent(cont_name, _id, parent_type):
+    cont = get_container(cont_name, _id)
+    if cont.get('parents') and cont['parents'].get(parent_type):
```
ehlertjd: The logic in this function seems slightly suspect - if my container doesn't have the parent it'll return the group instead.

api/dao/containerutil.py
```diff
@@ -64,8 +64,10 @@ def propagate_changes(cont_name, cont_ids, query, update, include_refs=False):
 
     if include_refs:
         analysis_query = copy.deepcopy(query)
+        analysis_update = copy.deepcopy(update)
+        analysis_update.pop('permissions', None)
         analysis_query.update({'parent.type': singularize(cont_name), 'parent.id': {'$in': cont_ids}})
-        config.db.analyses.update_many(analysis_query, update)
+        config.db.analyses.update_many(analysis_query, analysis_update)
```
ehlertjd: Does this mean we still have analyses out there with `permissions` on them?

hkethi002: The permissions key is being popped off the analysis_update a couple lines above

hkethi002: Although I think there's a bug in there

api/auth/containerauth.py
```diff
@@ -18,22 +18,24 @@ def f(method, _id=None, payload=None, unset_payload=None, recursive=False, r_pay
             errors = None
             if method == 'GET' and container.get('public', False):
                 has_access = True
+            elif not _check_scope(handler.scope, container, parent_container=target_parent_container):
```
nagem: What is this check covering that the method checks below are not?

api/auth/__init__.py
```diff
@@ -18,13 +18,26 @@
...
 
-def _get_access(uid, container):
+def _get_access(uid, container, scope=None):
+    if scope and _check_scope(scope, container):
+        return INTEGER_PERMISSIONS[scope['access']]
     permissions_list = container.get('permissions', [])
     for perm in permissions_list:
         if perm['_id'] == uid:
             return INTEGER_PERMISSIONS[perm['access']]
     return -1
```
nagem: I noticed that we directly call `_check_scope` a few places. Is that something that is required or could we have `_get_access` be the only place it is called? We could then remove the `if scope: <> else: True` logic.

api/auth/apikeys.py
```diff
@@ -131,8 +131,39 @@ def check(cls, api_key):
             raise APIAuthProviderException('Use of API key requires job to be in progress')
 
 
+class RuleApiKey(JobApiKey):
```
nagem: I think we would be able to reuse the `JobApiKey` class, just have an optional "scope" parameter on init.

api/web/base.py
```diff
@@ -72,6 +73,7 @@ def initialization_auth(self):
                     self.origin = {'type': Origin.user, 'id': self.uid}
```
nagem: I think the cleanest way to handle this would be to save the "origin" on all API keys. Right now, the SDK gear rule API keys will have an Origin "type" of `user` which isn't really true. The UI clients would likely try to link the "system" `_id` value to a value in the Users table and fail.

api/dao/hierarchy.py
```diff
@@ -39,6 +39,14 @@ def get_container(cont_name, _id):
         '_id': _id,
     })
 
+def get_parent(cont_name, _id, parent_type):
```
nagem: We already have `ContainerStorage.get_parent()`.  Could you add this as `get_parent_id()` to `ContainerStorage` instead?

nagem: If not, at least rename it to `get_parent_id`

api/jobs/handlers.py
```diff
@@ -46,17 +46,26 @@ class GearsHandler(base.RequestHandler):
...
+        gears = get_gears(all_versions=self.is_true('all_versions'))
+        if 'single_input' in filters:
+            filtered = True
+            gears = [gear for gear in gears if count_file_inputs(gear) <= 1]
+        if 'read_only_key' in filters:
+            filtered = True
+            gears = [gear for gear in gears if not requires_read_write_key(gear)]
+
+        if filtered:
+            return gears
```
nagem: I was worried how cleanly we would be able to write that double filter, looks good 👍

nagem:
I'm going to have to put some more thought into the API key change we made. Otherwise everything else has been re-reviewed.

nagem:
@kofalt release tag for gears lib has been updated.

api/jobs/handlers.py
```diff
@@ -478,11 +492,15 @@ def get_config(self, _id):
                     the_input = gear['gear']['inputs'][key]
 
                     if the_input['base'] == 'api-key':
-                        if j.origin['type'] != 'user':
+                        if j.origin['type'] == 'user':
+                            uid = j.origin['id']
+                            api_key = JobApiKey.generate(uid, j.id_)
+                        elif 'auto' in j.tags:
+                            project_id = cs_factory(pluralize(j.destination.type)).get_parent_id(j.destination.id, 'project')
+                            api_key = JobApiKey.generate(None, j.id_, scope={'level': 'project', 'id': project_id, 'access': 'ro'})
```
kofalt: This really looks like it should go inside a JobApiKey method.

Issue #1288, opened 2018-07-12
Pagination filter type coercion can be a little too aggressive


ehlertjd:
For example if I have a subject code of `1001`, I can't filter on `subject.code==1001` because it gets converted to a float.

PR #1378, opened 2018-08-27
Add regex operator to pagination filter

ehlertjd:
Allow regex filtering in pagination via `=~` operator.
e.g. `?filter=subject.code=~control.*`

Also is a workaround for #1288 

### Review Checklist

### DIFF
--- a/api/data_views/pipeline/filter.py
+++ b/api/data_views/pipeline/filter.py
@@ -1,5 +1,6 @@
 import bson
 import datetime
+import re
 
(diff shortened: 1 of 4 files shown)


nagem:
SGTM. We already have a few places that take regexes in the system (rules), but do you have any additional concerns we'd like to note with this one?

ehlertjd:
No major concerns, looking for feedback on the chosen syntax though. I think this is pretty straightforward and adds a ton of value to the already awesome pagination interface.

PR #1379, opened 2018-08-28
Fix packfile upload

davidfarkas:
To calculate the hash of the zip file we need to read the whole file after the zip is done. 
When we use OBS it uploads the file continuously during write operations. 
Since OBS cache is not implemented yet, we have to download the whole zipfile to calc its hash. 
The connection hangs during the download and the connection can be closed before the placer finished processing the zip.
Creating a zip with the same content, enable us to calculate the hash without downloading the zip.

### Review Checklist

### DIFF
--- a/api/files.py
+++ b/api/files.py
@@ -91,23 +91,20 @@ class FileProcessor(object):
 
         return form
 
-    def hash_file_formatted(self, filepath, f_system, hash_alg=None, buffer_size=65536):
+    @staticmethod
+    def hash_file_formatted(file_obj, hash_alg=None, buffer_size=65536):
         """
         Return the scitran-formatted hash of a file, specified by path.
         """
 
-        if not isinstance(filepath, unicode):
-            filepath = six.u(filepath)
-
         hash_alg = hash_alg or DEFAULT_HASH_ALG
         hasher = hashlib.new(hash_alg)
 
-        with f_system.open(filepath, 'rb') as f:
-            while True:
-                data = f.read(buffer_size)
-                if not data:
-                    break
-                hasher.update(data)
+        while True:
+            data = file_obj.read(buffer_size)
+            if not data:
+                break
+            hasher.update(data)
 
         return util.format_hash(hash_alg, hasher.hexdigest())
 
--- a/api/placer.py
+++ b/api/placer.py
@@ -3,13 +3,16 @@ import copy
 import datetime
 import dateutil
 import pymongo
+import os
 import uuid
 import zipfile
+
+from backports import tempfile
 
 import fs.path
 import fs.errors
 
-from . import config, util, validators
+from . import config, util, validators, files
 from .dao import containerutil, hierarchy
 from .dao.containerstorage import SessionStorage, AcquisitionStorage
 from .jobs import rules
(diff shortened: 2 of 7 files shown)


api/placer.py
```diff
@@ -514,6 +519,16 @@ def check(self):
         self.dir_ = util.sanitize_string_to_filename(self.a_label)
         self.name = self.dir_ + '.zip'
 
+        ######
+        # TODO: improve hash calculation
+        # Temp solution to avoid download the zip file again to calculate its hash
+        self.tempdir = tempfile.TemporaryDirectory(prefix='.tmp', dir=config.get_item('persistent', 'data_path'))
```
ambrussimon: Instead of the 3 variables, I suggest using something more compact:
* get rid of `self.tempdir` - grep says it's only used here
* `self.tmp_hash_zip = zipfile.ZipFile(tempfile.NamedTemporaryFile('w', ...), ...)`
* later, when `path` is needed for hash calculation: `open(self.tmp_hash_zip.filename, 'rb')`
* TODO comments need to be unified / more compact, too. I recommend:
```
\n
# TODO Improve hash calculation
line 1
[line2 ...]
\n
```

api/placer.py
```diff
@@ -563,15 +584,17 @@ def finalize(self):
             'type': str(Origin.user),
             'id': uid
         }
-
+        with open(self.local_path, 'rb') as f:
+            _hash = files.FileProcessor.hash_file_formatted(f)
+        os.remove(self.local_path)
```
ambrussimon: I _think_ the temp directory might linger around after this. Please make sure it doesn't.

api/files.py
```diff
@@ -91,23 +91,20 @@ def process_form(self, request, use_filepath=False):
...
         hash_alg = hash_alg or DEFAULT_HASH_ALG
         hasher = hashlib.new(hash_alg)
 
-        with f_system.open(filepath, 'rb') as f:
-            while True:
-                data = f.read(buffer_size)
-                if not data:
-                    break
-                hasher.update(data)
+        while True:
```
ambrussimon: I think a pre-read would be cleaner:
```
data = read()
while data:
    hasher.update(data)
    data = read()
````

nagem:
@roycollins @ehlertjd As mentioned on the Monday meeting, please see how this change is affected by the file storage layer PR. If it's fixed, go ahead and close this PR. Otherwise, please leave your recommendation on priority for addressing what this PR addresses.

roycollins:
I believe the intent of this PR is addressed in the file-abstraction-layer branch. 
It takes a slightly different approach in that the individual files are not copied directly to object storage.  Instead the individual files are saved on the local file system of the server and the zip file is also created on the local file system storage.  Only after the zip file is complete is it uploaded to object storage.  That eliminates the need to write the files to object storage and then read them back down to make the zip file. 

The assumption is made that the local filesystem will have ~2N space available.  We have to have at least 1N space to handle the file POST in either approach. So in this change we to also have at least as much space as the compressed file will consume, ~N.  

Ill keep it open until Justin has had a chance to review it as well to be sure Im not missing anything on the download side.

roycollins:
Justin and I both looked this over and we are pretty confident that this code is in the new file abstraction layer so I am going to close this PR.

PR #1381, opened 2018-08-28
Filter out deleted files when downloading

hkethi002:
Filters out deleted files when downloading and during the summary
### Review Checklist

### DIFF
--- a/api/download.py
+++ b/api/download.py
@@ -38,7 +38,7 @@ class Download(base.RequestHandler):
 
     def _append_targets(self, targets, cont_name, container, prefix, total_size, total_cnt, filters):
         inputs = [('input', f) for f in container.get('inputs', [])]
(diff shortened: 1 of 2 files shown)


ehlertjd:
Just want to check - Is there any chance of deleted containers being included? (e.g. delete an acquisition and download the project)

nagem:
@ehlertjd it looks like we will give you a deleted container if you ask for that specific container. We filter out deleted children of containers [example](https://github.com/flywheel-io/core/blob/master/api/download.py#L146).

But if you ask for a container directly, the deleted filter is not applied [example](https://github.com/flywheel-io/core/blob/master/api/download.py#L181).

nagem:
It's a little annoying, but I'd like us to merge this fix and open the second cleanup in a different PR.

PR #1389, opened 2018-09-04
Default modalities

### LINKED ISSUE #1384: Add additional default modalities
MR is currently the only modality we provide by default, along with it's classification system. Having additional modalities available, even without a full classification system, would be helpful to many customers.

Please add the following modalities:
CT, PET, US, EEG, iEEG, ~Genomics, Pathology~, X-ray, ECG, MEG, NIRS

cc @nagem @AshleyAncona  @jeffyager

hkethi002:
Fixes #1384 
New Modalities
### Review Checklist

### DIFF
--- a/bin/database.py
+++ b/bin/database.py
@@ -24,7 +24,7 @@ from api.types import Origin
 from api.jobs import batch
 
 
-CURRENT_DATABASE_VERSION = 57 # An int that is bumped when a new schema change is made
+CURRENT_DATABASE_VERSION = 58 # An int that is bumped when a new schema change is made
 
 
 def get_db_version():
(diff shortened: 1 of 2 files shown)


bin/database.py
```diff
@@ -1822,7 +1822,70 @@ def upgrade_to_54():
...
+def modality_maker():
+    modalities = [m['_id'] for m in config.db.modalities.find({})]
+    def upgrade_files_to_new_modalities(cont, cont_name):
+        files = cont.get('files', [])
+        for file_ in files:
+            if file_['classification'].get('Custom') and not file_.get('modality'):
+                if file_['classification']['Custom'][0] in modalities:
+                    file_['modality'] = file_['classification'].get('Custom')[0]
+        cont['files'] = files
+        config.db[cont_name].update_one({'_id': cont['_id']}, {"$set": {'files': files}})
```
ehlertjd: Suggest tracking whether or not anything has changed rather than letting the database figure that out. (i.e. only call update_one if at least one file was changed)

bin/database.py
```diff
@@ -1822,7 +1822,73 @@ def upgrade_to_54():
...
     process_cursor(cursor, upgrade_api_keys_to_54)
 
+def modality_maker():
+    modalities = [m['_id'] for m in config.db.modalities.find({})]
+    def upgrade_files_to_new_modalities(cont, cont_name):
+        files = cont.get('files', [])
+        file_updated = False
+        for file_ in files:
+            if file_['classification'].get('Custom') and not file_.get('modality'):
+                if file_['classification']['Custom'][0] in modalities:
```
nagem: I would turn this into a case-insensitive lookup.

bin/database.py
```diff
@@ -1822,7 +1822,73 @@ def upgrade_to_54():
...
+    }
+    ]
+
+    for modality in new_modalities:
+        if not config.db.modalities.find_one({'_id': modality['_id']}):
+            config.db.modalities.insert(modality)
 
+    closure = modality_maker()
+    for cont_name in ['projects', 'sessions', 'acquisitions', 'analyses']:
+        cursor = config.db[cont_name].find({'files.classification.Custom': {'$exists': True}})
```
nagem: You could limit the number of containers this has to iterate through by also looking for files that don't have a modality set.

nagem: You need a special query to make sure that you get containers with a file that has both a Custom classification as well as not having a modality, rather than containers that have 2+ files that match individually but don't have the union of the query. There should be an example somewhere in the recent db updates.

nagem:
Small suggestions but otherwise looks good 👍

bin/database.py
```diff
@@ -1822,7 +1822,79 @@ def upgrade_to_54():
...
+        for file_ in files:
+            if file_['classification'].get('Custom') and not file_.get('modality'):
+                try:
+                    modality_index = lower_modalities.index(file_['classification'].get('Custom')[0].lower())
+                    file_updated = True
+                    file_['modality'] = modalities[modality_index]
+                except ValueError:
+                    continue
+        if file_updated:
+            cont['files'] = files
```
ehlertjd: Shouldn't be necessary to re-assign the list of files as you've only changed the contents AFAICT

PR #1391, opened 2018-09-06
Add shell scripts to "source code"

lmperry:
A number of users are uploading shell scripts, which are not getting typed and cannot be viewed in the UI.

### Review Checklist

### DIFF
--- a/api/filetypes.json
+++ b/api/filetypes.json
@@ -25,7 +25,7 @@
     "log":          [ ".log" ],
     "pdf":          [ ".pdf" ],
     "presentation": [ ".ppt", ".pptx" ],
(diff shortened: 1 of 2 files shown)


nagem:
@lmperry Let me know if you want to re-type existing or just properly type all newly uploaded `.sh` files.

lmperry:
I don't know how common these are at this point, but it would be great to re-type the existing files. Thanks!

nagem:
@hkethi002 could you write a quick db update script? Just set it if it's not there.

nagem:
@hkethi002 this will need to get rebased FYI

PR #1395, opened 2018-09-12
Log gear info when job is started

### LINKED ISSUE #1393: Add job log containing gear version information when job state is changed to "running"
When a job is moved to "running", add a job log message giving information about the gear's version and name.

hkethi002:
Fixes #1393 
Name and Version are logged
### Review Checklist

### DIFF
--- a/api/jobs/queue.py
+++ b/api/jobs/queue.py
@@ -362,7 +362,8 @@ class Queue(object):
         # !!!
         # !!! DUPE WITH Queue.mutate
         # !!!
-        request = job.generate_request(get_gear(job.gear_id))
+        gear = get_gear(job.gear_id)
+        request = job.generate_request(gear)
 
         if peek:
             job.request = request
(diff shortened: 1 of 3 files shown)


api/jobs/queue.py
```diff
@@ -382,6 +383,8 @@ def start_job(tags=None, peek=False):
         if result is None:
             raise Exception('Marked job as running but could not generate and save formula')
 
+        Logs.add(job.id_, [{'msg': 'Gear Name: {}, Gear Version: {}\n'.format(gear['gear']['name'], gear['gear']['version']), 'fd': -1}])
```
kofalt: Okay interesting, I was assuming this whole time that this message would end up in the API logs, not the job logs, since at least for the user they can only get to these clicking through from the provenance tab where this info is already there.

I'd be fine with it here too, but please add an API log statement as well (that includes the gear and job IDs).

nagem: The main problem we are trying to solve is seeing a job log output stack after an error and having the first question be "what is the version of the gear?". API log statements are fine too, but there is not a way to quickly access uWSGI logs in a debug situation.

kofalt: Yeah, as long as it's in the API log as well I don't mind. I'd like to avoid bloating the job logs too much but this seems totally reasonable

PR #1404, opened 2018-09-21
delete -> delete_many

### LINKED ISSUE #1399: Inactivity timeout does not work
Issue seems to be related to comparing a number to a string (inactivity_timeout). https://github.com/flywheel-io/core/blob/master/api/web/base.py#L181

However, after casting inactivity_timeout to a number, this other error pops up that would also have to be addressed.
```
2018-09-18 12:12:28      scitran.api                  base.py   435:ERRO Traceback (most recent call last):
  File "/var/scitran/code/api/api/web/base.py", line 44, in __init__
    self.initialization_auth()
  File "/var/scitran/code/api/api/web/base.py", line 77, in initialization_auth
    self.uid = self.authenticate_user_token(session_token)
  File "/var/scitran/code/api/api/web/base.py", line 195, in authenticate_user_token
    config.db.refreshtokens.delete({'uid': cached_token['uid'], 'auth_type': cached_token['auth_type']})
  File "/usr/lib/python2.7/site-packages/pymongo/collection.py", line 3104, in __call__
    self.__name.split(".")[-1])
TypeError: 'Collection' object is not callable. If you meant to call the 'delete' method on a 'Collection' object it is failing because no such method exists.
 request_id=a58da473-1537272747
```

hkethi002:
Fixes #1399

Not sure if we should use delete_many or delete_one on the refreshtokens

### Review Checklist

### DIFF
--- a/api/web/base.py
+++ b/api/web/base.py
@@ -182,7 +182,7 @@ class RequestHandler(webapp2.RequestHandler):
 
                     # Token expired and no refresh token, remove and deny request
                     config.db.authtokens.delete_one({'_id': cached_token['_id']})
-                    config.db.refreshtokens.delete_many({'uid': cached_token['uid'], 'auth_type': cached_token['auth_type']})
+                    config.db.refreshtokens.delete_one({'uid': cached_token['uid'], 'auth_type': cached_token['auth_type']})
                     self.abort(401, 'Inactivity timeout')
 
                 # set last_seen to now
--- a/api/web/base.py
+++ b/api/web/base.py
@@ -178,11 +178,11 @@ class RequestHandler(webapp2.RequestHandler):
                 last_seen = cached_token.get('last_seen')
 
                 # If now - last_seen is greater than inactivity timeout, clear out session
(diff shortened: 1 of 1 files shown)


api/web/base.py
```diff
@@ -178,11 +178,11 @@ def authenticate_user_token(self, session_token):
                 last_seen = cached_token.get('last_seen')
 
                 # If now - last_seen is greater than inactivity timeout, clear out session
-                if last_seen and (timestamp - last_seen).total_seconds() > inactivity_timeout:
+                if last_seen and (timestamp - last_seen).total_seconds() > int(inactivity_timeout):
 
                     # Token expired and no refresh token, remove and deny request
                     config.db.authtokens.delete_one({'_id': cached_token['_id']})
-                    config.db.refreshtokens.delete({'uid': cached_token['uid'], 'auth_type': cached_token['auth_type']})
+                    config.db.refreshtokens.delete_many({'uid': cached_token['uid'], 'auth_type': cached_token['auth_type']})
```
ehlertjd: To me it looks like this should be `delete_one` since we only allow one instance of a refresh_token by uid/auth_type in the refreshtokens collection.

PR #1413, opened 2018-10-02
Treat empty hashes as non-matching for file replace

ehlertjd:
Seen at Columbia. Signed-url-uploads have no hash value, so all files were being treated as equal. As a result, reaper-replaced files were being ignored.

### Review Checklist


api/dao/hierarchy.py
```diff
@@ -269,6 +269,9 @@ def upsert_fileinfo(cont_name, _id, fileinfo, access_logger, ignore_hash_replace
 
     return container_before, container_after, saved_state
 
+def hashes_equal(hash1, hash2):
+    """Compare two hashes. They are equal if they both EXIST and are identical"""
+    return hash1 and hash2 and hash1 == hash2
```
kofalt: Suggest `hashes_equal_or_missing` for accuracy.

PR #1418, opened 2018-10-11
Ticket workflow for adding gears

kofalt:
This adds a ticket-based workflow for proposing and saving gears.
Used in conjunction with our custom image registry.

This adds four new routes, which may seem like a bit much. One is (designed to be) used exclusively by the registry authorizer, and later two old routes can be dropped.

Coming in a next commit will be the last addition to the save endpoint that ties this flow into that. For now, any ticket save will succeed.


api/jobs/handlers.py
```diff
@@ -73,6 +75,94 @@ def check(self):
         check_for_gear_insertion(self.request.json)
         return None
```
nagem: Was the `no-cover` for development purposes? I see no reason we can't have some integration tests around this workflow.

api/api.py
```diff
@@ -161,9 +161,20 @@ def prefix(path, routes):
             route('/<:[^/]+>/accept-failed-output', JobHandler,  h='accept_failed_output', m=['POST']),
         ]),
         route('/gears',                                  GearsHandler),
-        route('/gears/check',                            GearsHandler, h='check', m=['POST']),
-        route('/gears/temp',                             GearHandler, h='upload', m=['POST']),
-        route('/gears/temp/<cid:{oid}>',  GearHandler, h='download', m=['GET']),
+
+        # New upload flow
+        route('/gears/prepare-add',                      GearsHandler, h='prepare_add',     m=['POST']),
```
ehlertjd: Consider a more conventional URI scheme?
* `POST /gears/tickets` - Create ticket for gear creation
* `GET /gears/tickets` - Get own tickets 
* `GET /gears/tickets/<id>` - Get single ticket
* `POST /gears/tickets/<id>` - Save gear to database after upload

api/jobs/handlers.py
```diff
@@ -73,6 +75,94 @@ def check(self):
...
+        """
+
+        # Allow for just a summary of gear names
+        gear_names_only = self.is_true('gear_names_only')
+
+        result = config.db.gear_tickets.find({
+            'origin': self.origin
+        })
+
+        if gear_names_only:
```
ehlertjd: What's the use case for names only?

api/jobs/handlers.py
```diff
@@ -73,6 +75,94 @@ def check(self):
...
+    @require_admin
+    def save(self, _id): # pragma: no cover
+        """
+        Save a gear described by an upload ticket.
+        """
+
+        submit = self.request.json
+
+        ticket = config.db.gear_tickets.find_one({
+            '_id': bson.ObjectId(submit['ticket'])
```
ehlertjd: Raise 400/404 if `_id` is not a valid ObjectId

api/web/base.py
```diff
@@ -417,7 +417,8 @@ def handle_exception(self, exception, debug, return_json=False): # pylint: disab
             code = 500
 
         if code >= 400 and code < 500:
-            self.log.debug('client error: {}'.format(exception))
+            if 'elastic search not available: ConnectionError' not in str(exception):
```
ehlertjd: Guessing this shouldn't be here?

api/jobs/handlers.py
```diff
@@ -73,6 +75,94 @@ def check(self):
...
+        }
+
+    @require_admin
+    def get_ticket(self, _id): # pragma: no cover
+        """
+        Retrieve a gear-upload ticket.
+        """
+
+        result = config.db.gear_tickets.find_one({
+            '_id': bson.ObjectId(_id)
```
ehlertjd: Raise 400/404 if `_id` is not a valid ObjectId

api/jobs/handlers.py
kofalt: The registry authorization, which doesn't need any other information.

api/web/base.py
kofalt: Hmm, yeah this is something I've had in my local environment forever. This just squelches some ES flak when ES is not around. I'll remove it or move to a different PR

api/jobs/handlers.py
kofalt: We don't do that elsewhere in the API... are you wanting that to be a new standard? Maybe we should make a `try_object_id` util function.

api/jobs/handlers.py
kofalt: Duplicated from other sigs in the file. Most of this logic could have tests

api/api.py
kofalt: Yeah, that's cleaner.

PR #1420, opened 2018-10-12
Allow Read-Write Users to Move Sessions

LinaJohnson:
# Summary of changes
Decreased permission level on `PUT` for containers that have parents from admin to read-write. This allows for read-write users to move sessions from one project to another and to move session subjects as well. Related to a requirement for: https://github.com/flywheel-io/frontend/pull/1700

### DIFF
--- a/api/auth/containerauth.py
+++ b/api/auth/containerauth.py
@@ -43,9 +43,13 @@ def default_container(handler, container=None, target_parent_container=None):
                     errors = {'reason': 'permission_denied'}
 
             elif method == 'PUT' and target_parent_container is not None:
+                if target_parent_container.get('cont_name') == 'project' or target_parent_container.get('cont_name') == 'subject':
+                    required_perm = 'rw'
+                else:
+                    required_perm = 'admin'
                 has_access = (
-                    _get_access(handler.uid, container, scope=handler.scope) >= INTEGER_PERMISSIONS['rw'] and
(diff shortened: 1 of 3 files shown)


tests/integration_tests/python/test_containers.py
```diff
@@ -14,12 +14,20 @@ def test_switching_project_between_groups(data_builder, as_admin):
     assert r.json()['group'] == group_2
     assert r.json()['parents']['group'] == group_2
 
+    # Change permissions to read-write
+    user_id = as_user.get('/users/self').json()['_id']
+    assert as_admin.post('/projects/' + project + '/permissions', json={'_id': user_id, 'access': 'rw'}).ok
```
hkethi002: I think to test the switching permissions, as_user needs 'rw' permissions on the groups.

api/auth/containerauth.py
```diff
@@ -43,9 +43,13 @@ def f(method, _id=None, payload=None, unset_payload=None, recursive=False, r_pay
                     errors = {'reason': 'permission_denied'}
 
             elif method == 'PUT' and target_parent_container is not None:
+                if target_parent_container.get('cont_name') == 'project' or target_parent_container.get('cont_name') == 'subject':
```
nagem: I slightly cleaner way to write this would be 

```
if target_parent_container.get('cont_name') in ['project', 'subject', session']:
```

I also added session, to keep it consistent as "anything under project can be moved as a rw user.

LinaJohnson: @nagem done!

PR #1431, opened 2018-10-24
Job retry

hkethi002:
### Changes
- Wrote up some tests for the retry jobs endpoint
- Job retry endpoint goes through enqueue logic to generate the job object which we then modify
- Analysis jobs are updated to point to the latest job
  - If the analysis has outputs, superuser is required
- sdk gears use original user's api key
- job origin is the identical as well

### Review Checklist

### DIFF
--- a/api/jobs/handlers.py
+++ b/api/jobs/handlers.py
@@ -630,8 +630,19 @@ class JobHandler(base.RequestHandler):
                         j.inputs[x].check_access(self.uid, 'ro')
 
             j.destination.check_access(self.uid, 'rw')
+            if j.destination.type == 'analysis':
+                if len(j.destination.get().get('files', [])):
+                    raise APIPermissionException('Requires superuser to retry job if analysis has outputs')
 
-        new_id = Queue.retry(j, force=True)
+        # API key gear permission check
+        gear = get_gear(j.gear_id)
+        for x in gear['gear'].get('inputs', {}).keys():
+            input_ = gear['gear']['inputs'][x]
+            if input_.get('base') == 'api-key':
+                if not self.user_is_admin and self.uid != j.origin['id']:
+                    raise APIPermissionException('Only original scheduler or root user can retry a gear requiring an api key input')
+
+        new_id = Queue.retry(j, force=True, only_failed=not self.is_true('ignoreState'))
         return { "_id": new_id }
 
     @require_drone
--- a/api/jobs/handlers.py
+++ b/api/jobs/handlers.py
@@ -622,7 +622,7 @@ class JobHandler(base.RequestHandler):
         j = Job.get(_id)
 
         # Permission check
-        if not self.superuser_request:
+        if not self.user_is_admin:
 
             if j.inputs is not None:
                 for x in j.inputs:
--- a/api/jobs/queue.py
+++ b/api/jobs/queue.py
@@ -109,9 +109,9 @@ class Queue(object):
 
         if job.state in ['cancelled', 'complete']:
             if only_failed:
-                raise Exception('Can only retry a job that is failed, please use only_failed parameter')
+                raise InputValidationException('Can only retry a job that is failed, please use only_failed parameter')
         elif job.state != 'failed':
-            raise Exception('Can not retry running or pending job')
+            raise InputValidationException('Can not retry running or pending job')
 
 
         if job.request is None:
--- a/api/jobs/queue.py
+++ b/api/jobs/queue.py
@@ -144,7 +144,7 @@ class Queue(object):
 
         result = config.db.jobs.update_one({"_id": bson.ObjectId(job.id_)}, {'$set': {"retried": new_job.created}})
         if result.modified_count != 1:
-            raise Exception('Could not set retried time for job {}'.format(job.id_))
+            log.error('Could not set retried time for job {}'.format(job.id_))
 
         new_id = new_job.insert(ignore_insertion_block=True)
         log.info('respawned job %s as %s (attempt %d)', job.id_, new_id, new_job.attempt)
--- a/api/jobs/queue.py
+++ b/api/jobs/queue.py
@@ -97,7 +97,7 @@ class Queue(object):
             Queue.retry(job)
 
     @staticmethod
-    def retry(job, force=False):
+    def retry(job, force=False, only_failed=True):
         """
         Given a failed job, either retry the job or fail it permanently, based on the attempt number.
         Can override the attempt limit by passing force=True.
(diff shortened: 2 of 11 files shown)


api/jobs/handlers.py
```diff
@@ -630,8 +630,11 @@ def retry(self, _id):
                         j.inputs[x].check_access(self.uid, 'ro')
 
             j.destination.check_access(self.uid, 'rw')
+            if j.destination.type == 'analysis':
+                if len(j.destination.get().get('files', [])):
+                    raise APIPermissionException('Requires superuser to retry job if anlaysis has outputs')
```
ehlertjd: spelling (analysis)

api/jobs/queue.py
```diff
@@ -108,7 +108,11 @@ def retry(job, force=False):
             return
 
         if job.state != 'failed':
-            raise Exception('Can only retry a job that is failed')
+            if only_failed:
+                raise Exception('Can only retry a job that is failed, please use force parameter')
+            elif job.state not in ['cancelled', 'complete']:
+                raise Exception('Can only retry a job that is not mutating')
```
ehlertjd: Maybe "Cannot retry a running or pending job?"

api/jobs/queue.py
```diff
@@ -108,7 +108,11 @@ def retry(job, force=False):
             return
 
         if job.state != 'failed':
-            raise Exception('Can only retry a job that is failed')
+            if only_failed:
+                raise Exception('Can only retry a job that is failed, please use force parameter')
```
ehlertjd: "please override the only_failed parameter"

api/jobs/queue.py
```diff
@@ -108,7 +108,11 @@ def retry(job, force=False):
             return
 
         if job.state != 'failed':
-            raise Exception('Can only retry a job that is failed')
+            if only_failed:
```
ehlertjd: I would switch the order of these checks for user experience
(i.e. try to re-run a pending job and get the parameter error, then try setting that and get an error regarding state)

api/jobs/queue.py
```diff
@@ -121,21 +125,24 @@ def retry(job, force=False):
...
-        now = datetime.datetime.utcnow()
-        new_job.created = now
-        new_job.modified = now
+        new_job.request = copy.deepcopy(job.request)
+        new_job.id_ = bson.ObjectId()
 
         # update input uris that reference the old job id
-        for i in new_job.request['inputs']:
+        for i in new_job.request['inputs']+new_job.request['outputs']:
+            config.log.debug(i)
```
ehlertjd: remove debug statement?

tests/integration_tests/python/test_analyses.py
```diff
@@ -119,13 +119,59 @@ def test_online_analysis(data_builder, as_admin, as_drone, file_form, api_db):
     r = as_admin.post('/analyses/' + analysis + '/files', files=file_form('output.csv'))
     assert r.status_code == 400
 
-    # Engine upload
+    # # Engine upload
```
ehlertjd: ?

tests/integration_tests/python/test_jobs.py
```diff
@@ -1276,9 +1259,192 @@ def test_scoped_job_api_key(randstr, data_builder, default_payload, as_public, a
...
+    # get config
+    r = as_root.get('/jobs/' + retried_job_id + '/config.json')
+    assert r.ok
+    config = r.json()
+
+    assert config['destination']['id'] == acquisition
+    assert type(config['config']) is dict
+    retried_api_key = config['inputs']['api_key']['key']
+    # ensure api_key works
+    as_retried_job_key = as_public
```
ehlertjd: This just creates a reference to `as_job_key` - generally, I would just get rid of `as_retried_job_key` since it's the same as `as_job_key`

tests/integration_tests/python/test_jobs.py
```diff
@@ -1276,9 +1259,192 @@ def test_scoped_job_api_key(randstr, data_builder, default_payload, as_public, a
...
-    r = as_root.put('/jobs/' + job_id, json={'state': 'complete'})
+    # fail job and ensure API key no longer works
+    r = as_root.put('/jobs/' + job_id, json={'state': 'failed'})
     assert r.ok
 
     r = as_job_key.get('/projects/' + project)
     assert r.status_code == 401
+
+    # Retry job
+    r = as_admin.post('/jobs/' + job_id + '/retry')
```
ehlertjd: Test retries as user who scheduled job, and as another r/w, non-admin user?

api/jobs/handlers.py
```diff
@@ -630,8 +630,11 @@ def retry(self, _id):
                         j.inputs[x].check_access(self.uid, 'ro')
 
             j.destination.check_access(self.uid, 'rw')
```
ehlertjd: efficiency: `check_access` appears to unconditionally call `get`, which we then (possibly) do again below

api/jobs/queue.py
```diff
@@ -212,6 +219,8 @@ def enqueue_job(job_map, origin, perm_check_uid=None):
         destination = None
         if job_map.get('destination', None) is not None:
             destination = create_containerreference_from_dictionary(job_map['destination'])
+            # Check that it exists
+            destination.get()
```
ehlertjd: Similar to above, we could end up retrieving destination from the database 2x (get and check_access)

api/jobs/handlers.py
hkethi002: If it is a superuser request, we skip over that entire block of checking access. Both here and later in the enqueue logic.

ehlertjd:
Could you also add the query parameters to swagger docs?

api/jobs/queue.py
```diff
@@ -107,8 +107,12 @@ def retry(job, force=False):
             log.info('Permanently failed job %s (after %d attempts)', job.id_, job.attempt)
             return
 
-        if job.state != 'failed':
-            raise Exception('Can only retry a job that is failed')
+        if job.state in ['cancelled', 'complete']:
+            if only_failed:
+                raise Exception('Can only retry a job that is failed, please use only_failed parameter')
```
ehlertjd: I think this (and below) should be InputValidation errors (i.e. HTTP status code 400 instead of 500)

hkethi002:
Realized that I got my comments mixed up and rebased the wrong branch/repo. Hopefully the second commit isn't too bad of a diff.

ehlertjd:
One final thought occurred to me @hkethi002 & @nagem  - Can we set a `retried` timestamp on jobs that have been retried? I think that would improve CLI user experience (i.e. don't present jobs that cannot be retried because they have already)

api/jobs/queue.py
```diff
@@ -140,7 +140,11 @@ def retry(job, force=False, only_failed=True):
         if new_job.destination.type == 'analysis':
             config.db.analyses.update_one({'_id': bson.ObjectId(new_job.destination.id)},
                                           {'$set': {'job': str(new_job.id_),
-                                                    'modified': datetime.datetime.utcnow()}})
+                                                    'modified': new_job.created}})
+
+        result = config.db.jobs.update_one({"_id": bson.ObjectId(job.id_)}, {'$set': {"retried": new_job.created}})
+        if result.modified_count != 1:
+            raise Exception('Could not set retried time for job {}'.format(job.id_))
```
ehlertjd: maybe just `log.error`

PR #1438, opened 2018-10-29
Config for upload limit

hkethi002:
### Review Checklist


api/config.py
```diff
@@ -47,7 +47,8 @@
         'central_url': 'https://sdmc.scitran.io/api',
         'registered': False,
         'ssl_cert': None,
-        'inactivity_timeout': None
+        'inactivity_timeout': None,
+        'upload_limit': 10
```
ehlertjd: 10 what? Can we include units?

hkethi002:
Realized that env vars are strings so the config value will be a string. Not sure if this is the best way, but other values that were integers were also strings and so there wasn't a good way to make the upload a string without changing those value types to integers as well.

PR #1440, opened 2018-10-31
Only create subject if switching to new project

hkethi002:
Fixes the UNC problem of creating duplicate subjects
### Review Checklist

### DIFF
--- a/api/handlers/containerhandler.py
+++ b/api/handlers/containerhandler.py
@@ -404,7 +404,7 @@ class ContainerHandler(base.RequestHandler):
             # Handle changing project (moving session and subject to another project)
             # * Copy subject into target project if there are other sessions on it
             # * Move if this is the only session on it (else branch)
-            elif target_project and config.db.sessions.count({'subject': container['subject']['_id']}) > 1:
+            elif (target_project and project_id != current_project['_id']) and config.db.sessions.count({'subject': container['subject']['_id']}) > 1:
                 subject = copy.deepcopy(container['subject'])
                 subject.pop('parents')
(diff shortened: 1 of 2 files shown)


api/handlers/containerhandler.py
```diff
@@ -404,7 +404,7 @@ def put(self, cont_name, **kwargs):
             # Handle changing project (moving session and subject to another project)
             # * Copy subject into target project if there are other sessions on it
             # * Move if this is the only session on it (else branch)
-            elif target_project and config.db.sessions.count({'subject': container['subject']['_id']}) > 1:
+            elif (target_project and project_id != current_project['_id']) and config.db.sessions.count({'subject': container['subject']['_id']}) > 1:
```
nagem: Above we have this if statement:
https://github.com/flywheel-io/core/pull/1440/files#diff-90f25ae1793437dd630278372aa93828R385

How are these two related? Can we ensure the subject and session get moved (and not recreated) if there is only 1 session?

nagem: @ambrussimon do you have any insight here? I might just be reading this incorrectly.

hkethi002: We set the subject payload id to the current subject in the following if statement. I believe we test this case here: https://github.com/flywheel-io/core/pull/1440/files#diff-beb22bc647a4925ff634a121502d4032R348

ambrussimon: Oh boy this is ugly. @nagem:

> How are these two related?

* 1st if is _test collision and bail early if so_
* 2nd is _OK no collision, but if moving, check how many sessions there are_
(because if more, then do not move tree for safety reasons.)

PR #1441, opened 2018-11-01
Add swagger for job completion statistics

ehlertjd:
These are the **Proposed** statistics to collect for job execution.

Request for comments!

Adds additional metadata around jobs, including completion stats.

This is useful to us because it lets us answer questions about billing (i.e. what group used what resources for how long).
It also lets us build up a knowledge base for cloud-scale tuning and scheduling, as well as offering some insight for job failures (e.g. if jobs running on 2 cpus with low memory are failing, etc)

How attributes will be populated:
- group & project are set based on destination container at job creation
- started is set when the job is transitioned to running
- completed is set when the job is transitioned to complete or failed
- failure_reason comes from engine, when it marks a job as failed

- job-stats
  - instance-info - comes from engine, preferably on job start
  - version-info - comes from core/engine, preferably on job start
  - input/output files & sizes - come from core, at job start & end
  - preparation time - comes from engine, at execution start
  - execution time - comes from engine, at execution end
  - upload time - comes from engine, post execution
  - total time - calculated as (end time - start time)

### Review Checklist

### DIFF
--- a/swagger/schemas/definitions/job.json
+++ b/swagger/schemas/definitions/job.json
@@ -127,6 +127,85 @@
         }
       }
     },
+    "version-info": {
+      "type": "object",
+      "additionalProperties": {
+          "type": "string"
+      },
+      "description": "An object that contains known version info for components involved in the job"
+    },
+    "instance-info": {
+      "type": "object",
+      "properties": {
+          "name": {
+            "type": "string",
+            "description": "A machine or instance name, possibly a FQDN"
+          },
+          "instance_type": {
+            "type": "string",
+            "description": "A free form string describing the instance type"
+          },
+          "cpu_cores": {
+            "type": "integer",
+            "description": "The number of CPU cores"
+          },
+          "gpu": {
+            "type": "boolean",
+            "description": "Whether or not a GPU is available"
+          },
+          "memory_bytes": {
+            "type": "integer",
+            "description": "The amount of memory on the system, in bytes"
+          },
+          "disk_bytes": {
+            "type": "integer",
+            "description": "The size of the hard disk on the system, in bytes"
+          },
+          "swap_bytes": {
+            "type": "integer",
+            "description": "The available swap space, in bytes"
+          }
+        }
+    },
+    "job-stats": {
+      "type": "object",
+      "properties": {
+          "total_input_files": {
+            "type": "integer",
+            "description": "The number of input files"
+          },
+          "total_input_size_bytes": {
+            "type": "integer",
+            "description": "The combined size of all input files, in bytes"
+          },
+          "total_output_files": {
+            "type":"integer",
+            "description": "The number of output files"
+          },
+          "total_output_size_bytes": {
+            "type": "integer",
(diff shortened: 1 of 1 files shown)


ehlertjd:
@gsfr Looking for feedback on whether there's anything missing or extraneous.

nagem:
SGTM, I couldn't think of anything else I would add that we haven't already talked through. I like the units being part of the key name.

gsfr:
Looks comprehensive to me. I would drop the `-info` postfix on a couple of the property names. Also, the use of dash vs underscore is unclear to me. For what it's worth, I have come to the conclusion that JSON [property names should be expressed in camelCase](https://google.github.io/styleguide/jsoncstyleguide.xml?showone=Property_Name_Format#Property_Name_Format).

swagger/schemas/definitions/job.json
```diff
@@ -127,6 +127,89 @@
...
+            "type": "string",
+            "description": "The hostname or (more likely) IP address of the engine instance"
+          },
+          "instance_type": {
+            "type": "string",
+            "description": "A free form string describing the instance type"
+          },
+          "cpu_cores": {
+            "type": "integer",
+            "description": "The number of CPU cores"
```
kofalt: Physical or virtual?

swagger/schemas/definitions/job.json
```diff
@@ -127,6 +127,89 @@
...
+            "type": "integer",
+            "description": "The length of time that the job took to execute, in seconds"
+          },
+          "upload_time_sec": {
+            "type": "integer",
+            "description": "The length of time taken to upload the job's outputs, in seconds"
+          },
+          "total_time_sec": {
+            "type": "integer",
+            "description": "The total length of time from start to finish, in seconds"
```
kofalt: Pending to complete/cancelled/failed?

ehlertjd: Running to complete/failed. Likely not recorded if cancelled.

hkethi002:
I know initially we were talking about making this its own collection, is there any reason we moved away from that?

kofalt:
I don't have a strong opinion on putting stats in a separate collection, but the jobs object never ceases to grow, and it might make aggregate queries cleaner / faster. So gentle :+1: from me there.

This looks like a good batch of data points. I would like to add some memory consumption / core consumption stuff, but we can leave that for another time. I would stress that each value in the stats object is optional.

When I designed transition timestamps in my mind, I took a different approach:

```js
'transition': {   // idk, whatever name you want
    'pending': time,
    'running': time,
    ...
}
```

With the implementation being that whenever the object moves into state `X`, then `transition.X` is set to `time.now`. This IMO is a bit cleaner than separate state keys on the top-level object.

I would support moving the stats and transition-ish objects to a new object on a separate collection.

kofalt:
Honestly, I also hesitate to :+1: group / project keys relative to a relevant-containers array. Let's talk more later this week.

ehlertjd:
Closing in favor of implementation PR: #1446

PR #1442, opened 2018-11-05
Engine metadata tags

### LINKED ISSUE #1422: Allow the setting of tags via engine upload metadata
Tags are not currently an allowable field in the metadata input json schema for the engine. There has been a customer request to allow this behavior, specifically for classification gears that wish to tag containers (sessions) based on information in file headers. A potential list of containers to allow this behavior would be:
  - [x] file
  - [x] acquisition
  - [x] session
  - [x] subject
  - [x] project

Open questions:
  - How to handle tags that are not in the group's tag pool?
    - Rejecting the entire engine result (400) seems extreme. Maybe just remove tags that are not allowed? That behavior might not be expected by the user and go unnoticed.
  - Does is create a union of existing tags and those provided in the metadata or replace existing tags?
    - Create a union

LinaJohnson:
# Resolves #1422 

# Description
Tags are now an allowable field in metadata input JSON schemas for engine uploads. The containers covered are:
- file
- acquisition
- session
- subject
- project

### DIFF
--- a/tests/integration_tests/python/test_uploads.py
+++ b/tests/integration_tests/python/test_uploads.py
@@ -17,7 +17,8 @@ def upload_file_form(file_form, merge_dict, randstr):
         meta = {
             'project': {
                 'label': prefix + '-project-label',
-                'files': [{'name': files['project']}]
+                'files': [{'name': files['project']}],
+                'tags': ['one', 'two']
             },
             'session': {
                 'uid': prefix + '-session-uid',
@@ -25,12 +26,14 @@ def upload_file_form(file_form, merge_dict, randstr):
                 'subject': {
                     'code': prefix + '-subject-code'
                 },
-                'files': [{'name': files['session']}]
+                'files': [{'name': files['session']}],
+                'tags': ['one', 'two']
             },
             'acquisition': {
                 'uid': prefix + '-acquisition-uid',
                 'label': prefix + '-acquisition-label',
-                'files': [{'name': files['acquisition']}]
+                'files': [{'name': files['acquisition']}],
+                'tags': ['one', 'two']
             }
         }
         if meta_override:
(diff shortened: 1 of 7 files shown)


nagem:
This one will end up being a little more complicated than allowing the setting of labels through the input schemas. We would want any tags that are specified in the list to be added to the existing tags list on the object. So if the project looked like:

```
{
    "label": "A Project",
    "tags": ["one", "two"],
    ... 
}
```

An engine upload request with:
```
{
    "project": {
        "tags": ["two", "three"]
    }
}
```

Should result in a project that looks like this:
```
{
    "label": "A Project",
    "tags": ["one", "two", "three"],
    ... 
}
```

This PR as it stands would result in the tags list being _replaced_ by the update, and not a unique union of the two tag lists. 

What we would need to do to get the expected outcome:
1. Use [$addToSet](https://docs.mongodb.com/manual/reference/operator/update/addToSet/) for all tag updates to containers caused by engine upload. 
2. Update/Add tests to show that tags are added to the list when they are not already in the list, ignored otherwise.

nagem:
For some help finding the relevant container updating logic:

This is where the metadata is applied:
https://github.com/flywheel-io/core/blob/master/api/placer.py#L371

Followed by this call:
https://github.com/flywheel-io/core/blob/master/api/dao/hierarchy.py#L555

Which eventually gets us to a bulk update call:
https://github.com/flywheel-io/core/blob/master/api/dao/hierarchy.py#L615

That for loop will need to have a special if statement for when the key (`k`) is `tags`, and do an `$addToSet` call instead of a regular `$set` call, I believe.

tests/integration_tests/python/test_uploads.py
```diff
@@ -1486,3 +1498,52 @@ def test_packfile_upload(data_builder, file_form, as_admin, as_root, api_db):
...
+            ],
+            'tags': ['one', 'two']
+        }
+    }
+
+    r = as_root.post('/engine',
+        params={'level': 'project', 'id': project, 'filename_path':True},
+        files=file_form('one.csv', 'two.csv', 'folder/three.csv', meta=metadata)
+    )
+    assert r.ok
```
nagem: I'd like to see another engine upload here to assert that the `$addToSet` is working as expected. No need to upload files, just a simple metadata engine upload that sets this project's tags to `['two', 'three']` and confirm that the project's tags afterwards are `['one', 'two', 'three']`, or "two was ignored, three was added".

nagem:
After the one testing change this LGTM!

PR #1445, opened 2018-11-06
Generate parents for subjects created through uploads

### LINKED ISSUE #1444: Parents sub-doc does not get populated in label/reaper uploads
Likely affects SDK gear rules permission checks.

Added [checks in integration tests](https://github.com/flywheel-io/core/commit/f24e35a3652930950157129e176801ba780ccf02) on the `no-parents` branch to illustrate the problem.

We will need to fix this, and add an upgrade script to correct the affected files.

hkethi002:
Fixes #1444 
### Review Checklist

### DIFF
--- a/bin/database.py
+++ b/bin/database.py
@@ -24,7 +24,7 @@ from api.types import Origin
 from api.jobs import batch
 
 
-CURRENT_DATABASE_VERSION = 56 # An int that is bumped when a new schema change is made
+CURRENT_DATABASE_VERSION = 57 # An int that is bumped when a new schema change is made
 
 
 def get_db_version():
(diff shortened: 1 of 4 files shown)


ehlertjd:
Still probably need a db-upgrade to fix existing reaper uploads.

nagem:
It looks like only those on EM5 (4.2.x) should be affected by this bug, correct?

bin/database.py
```diff
@@ -1931,6 +1931,41 @@ def upgrade_to_56():
...
+    CHILD_MAP ={
+        "subjects": "sessions",
+        "sessions": "acquisitions"
+    }
+
+    cont_type = containerutil.singularize(cont_name)
+
+    # Handle subject without parents
+    if cont_type == 'subject':
+        subject_group = config.db.projects.find_one({'_id': cont['project']})
```
nagem: A little confusingly named, but since it's a db update that's a low priority.

bin/database.py
```diff
@@ -1931,6 +1931,41 @@ def upgrade_to_56():
...
+            group_id = subject_group['group']
+        parents = {'project': cont['project'], 'group': group_id}
+        config.db.subjects.update({'_id': cont['_id']}, {'$set': {'parents': parents}})
+    else:
+        parents = cont.get('parents', {})
+
+    parents[cont_type] = cont['_id']
+    config.db['analyses'].update_many({'parent.id': cont['_id']}, {'$set': {'parents': parents}})
+
+    if cont_name != 'acquisitions':
```
nagem: Why would cont_name ever be 'acquisitions'? It looks like upgrade_to_57 is only called on subjects.

nagem: My previous comment was incorrect, I see it is recursively updating all children here.

PR #1457, opened 2018-11-15
Modify upgrade 55 and 60

hkethi002:
Keeps relevant top level subject keys on the session.
### Review Checklist

### DIFF
--- a/api/dao/containerutil.py
+++ b/api/dao/containerutil.py
@@ -113,6 +113,13 @@ def extract_subject(session, project):
     session['subject'] = subject['_id']
     if subject.get('age'):
         session['age'] = subject.pop('age')
+    raw_subject_fields = ['firstname', 'lastname', 'sex', 'race', 'ethnicity']
+    subject_raw = {k:v for k,v in subject.iteritems() if v is not None and k in raw_subject_fields}
+    if subject_raw:
+        if session.get('info'):
+            session['info']['subject_raw'] = subject_raw
+        else:
+            session['info'] = {'subject_raw': subject_raw}
     return subject
 
 
--- a/api/dao/containerutil.py
+++ b/api/dao/containerutil.py
@@ -79,6 +79,18 @@ def propagate_changes(cont_name, cont_ids, query, update, include_refs=False):
             propagate_changes(child_cont, child_ids, query, update, include_refs=include_refs)
 
 
+def attach_raw_subject(session, subject, additional_fields=None):
+    raw_subject_fields = ['firstname', 'lastname', 'sex', 'race', 'ethnicity']
+    if additional_fields:
+        raw_subject_fields += additional_fields
+    subject_raw = {k: copy.deepcopy(v) for k, v in subject.iteritems() if v is not None and k in raw_subject_fields}
+    if subject_raw:
+        if session.get('info'):
+            session['info']['subject_raw'] = subject_raw
+        else:
+            session['info'] = {'subject_raw': subject_raw}
+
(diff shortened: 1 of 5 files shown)


bin/database.py
```diff
@@ -1847,6 +1847,17 @@ def extract_subject(session):
         if subject.get('age'):
             session['age'] = subject.pop('age')
         session['subject'] = subject['_id']
+        subject_raw = {
+            'firstname': subject.get('firstname'),
+            'lastname': subject.get('lastname'),
+            'race': subject.get('race'),
+            'ethnicity': subject.get('ethnicity')
+        }
+        subject_raw = {k:v for k,v in subject.iteritems() if v is not None}
```
ehlertjd: This makes line 1850-1855 redundant, right?

bin/database.py
```diff
@@ -2118,6 +2130,17 @@ def extract_subject(session):
...
```
ehlertjd: This makes line 2133-2138 redundant, right?

bin/database.py
```diff
@@ -1864,7 +1875,8 @@ def merge_dict(a, b):
                 merge_dict(a[k], b[k])
             else:  # handle conflict
                 logging.warning('merge conflict on key %s on subject %s', k, a.get('_id') or b.get('_id'))
-                a.setdefault(k + '_history', []).append(a[k])
+                if k not in ['firstname', 'lastname', 'race', 'ethnicity']:
```
ehlertjd: Why not always set this?

bin/database.py
```diff
@@ -2209,13 +2233,13 @@ def merge_dict(a, b):
 
 
 
-        cursor = config.db.subjects.find({'created': None})
-        logging.info("Adding in created timestamps for subjects")
-        process_cursor(cursor, add_subject_created_timestamps)
+    cursor = config.db.subjects.find({'created': None})
+    logging.info("Adding in created timestamps for subjects")
+    process_cursor(cursor, add_subject_created_timestamps)
```
ehlertjd: 🎉

api/dao/containerutil.py
```diff
@@ -113,6 +113,13 @@ def extract_subject(session, project):
     session['subject'] = subject['_id']
     if subject.get('age'):
         session['age'] = subject.pop('age')
+    raw_subject_fields = ['firstname', 'lastname', 'sex', 'race', 'ethnicity']
+    subject_raw = {k:v for k,v in subject.iteritems() if v is not None and k in raw_subject_fields}
+    if subject_raw:
+        if session.get('info'):
+            session['info']['subject_raw'] = subject_raw
+        else:
+            session['info'] = {'subject_raw': subject_raw}
```
ehlertjd: Maybe turn this block into a helper function `set_subject_raw` in `containerutil`? `database.py` appears to import containerutil.

tests/integration_tests/python/test_uploads.py
```diff
@@ -6,6 +6,15 @@
 import pytest
 
 
+def get_full_container(user, url, index):
+    _, _, _, cont_type = url.split('/')
```
nagem: Mind adding a comment about the goal of this helper function? It took me a minute to realize it was because you don't know the id of the container. 

Also, you can use `url.split('/')[-1]` to get the last item in a split.

nagem: Or you can start arguing with SO people about if the single digit increase in speed percent is worth the confusion around `str.rsplit()`.

PR #1459, opened 2018-11-26
Update data view column definitions for subjects

ehlertjd:
Adds new columns for subject `type`, `cohort`, `species` and `strain`.
Adds `weight` column to session.
Moves `age` columns from subject to session.

### Review Checklist


nagem:
Are you worried about existing views that `reference subject.age`?

ehlertjd:
@nagem That's a tough question, and I didn't have a great answer other than letting existing views "break" (with an empty `subject.age`) in this instance.

We could alias `subject.age` to `session.age` for example, but then a subject retrieval would inconspicuously result in retrieving sessions as well, which is probably an undesired side-effect in most cases.

ehlertjd:
@nagem This is no longer a breaking change. I kept the subject age columns, but they won't be listed by the endpoint.

PR #1462, opened 2018-11-28
Add login count to metrics

hkethi002:
The db metrics also use max instead of live sum to prevent the different workers from adding up their results.
### Review Checklist


api/metrics/collect.py
```diff
@@ -91,14 +92,47 @@ def collect_db_metrics():
...
+
         # Get gear versions
         gear_count = 0
         for gear_doc in config.db.gears.find():
             gear = gear_doc.get('gear', {})
             name = gear.get('name', 'UNKNOWN')
             version = gear.get('version', 'UNKNOWN')
             created = str(gear_doc.get('created', 'UNKNOWN'))
-            values.GEAR_VERSIONS.labels(name, version, created).set(1)
+            count = config.db.jobs.find({'gear_id': str(gear_doc['_id'])}).count()
```
ehlertjd: Would recommend using aggregation framework to do this in one database query above, to create a lookup map for setting these counts, rather than N queries. (i.e. groupBy gear_id and sum 1 per job)

api/metrics/collect.py
```diff
@@ -91,14 +92,47 @@ def collect_db_metrics():
             count = config.db[collection_name].count()
             values.COLLECTION_COUNT.labels(collection_name).set(count)
 
+        # Get access logs of type user login
+        login_counts = config.log_db.access_log.inline_map_reduce(
```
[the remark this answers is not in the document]
hkethi002: It's about 4x slower than a regex '@(?!flywheel)' filter on find, I'll change it to that

api/metrics/collect.py
```diff
@@ -125,12 +116,20 @@ def collect_db_metrics():
...
+                }
+            }
+        ])}
+        gears = config.db.find()
+        for gear in gears:
+            name = gear['_id'].get('name', 'UNKNOWN')
+            version = gear['_id'].get('version', 'UNKNOWN')
             created = str(gear_doc.get('created', 'UNKNOWN'))
-            count = config.db.jobs.find({'gear_id': str(gear_doc['_id'])}).count()
+            count = job_count_by_gear[gear['_id']]
```
ehlertjd: `job_count_by_gear.get(gear['_id']], 0)`?

nagem:
> Any concern with explicit flywheel.io domain in core code?

We are now flywheel-io/core so I would say no.

PR #1466, opened 2018-12-04
Populate and maintain parents on jobs

ehlertjd:
Rather than group/project, we set the parents object to the destination's parents (plus destination).
This allows us to maintain the parents when containers are moved, and always add any touched containers to the `related_container_ids` collection.

In short:
- `parents` will always accurately reflect a job's destination's parents, as long as the destination is valid
- `related_container_ids` is a set of all containers associated with a job, over time. Nothing gets removed from that set.

This also includes an upgrade step to populate parents for existing jobs.

### Review Checklist


ehlertjd:
`parents` is tied directly to the destination, therefore easy to maintain consistency when the destination container moves. The change was made from group/project to `parents` to facilitate easy updates when a container is moved. Destination (i.e. `parents`) is what will be used to attribute a job to a particular group for billing / scheduling purposes. AFAIK, there's nothing that says there couldn't be multiple groups in `related_container_ids`, which makes it unsuitable for that purpose.

By virtue of being a set that also includes inputs, `related_container_ids` is impossible to update with a one-shot update after a move - we would need to iterate over each job that includes that container id in the `related_container_ids` and re-calculate `related_container_ids`. Given that moves should be rare, maybe we could/should take the hit and recalculate all related jobs, but I was a little worried about the complexity - expense of such an operation. 

@kofalt

PR #1468, opened 2018-12-05
Add auth/status endpoint

ehlertjd:
Peripheral services such as perimeter and registry token service need an endpoint where an API key can be checked for validity.

Simply put, this endpoint will return a `200` response with the origin and user_is_admin values from the request if the API key is valid, otherwise will return an appropriate status code (e.g. `403`)

### Review Checklist


nagem:
I like it 👍

PR #1469, opened 2018-12-06
Verify gear images with ticket workflow

kofalt:
A revised take on a more robust gear verification workflow. Gear additions now use a ticketing workflow that is validated against the docker registry, rather than just using bona-fides.

As usual, editor caught some whitespace; add `?w=1` to diff URL as desired.


api/jobs/gears.py
```diff
@@ -222,3 +224,58 @@ def check_for_gear_insertion(doc):
...
+    """
+    Format config values for registry authentication.
+    """
+
+    c = config.get_config()
+    pw = c['core']['drone_secret']
+    url =  urlparse(c['site']['redirect_url'])
+    host = url.hostname + ':' + str(url.port)
+    api_key = ''
+    un = 'device.flywheel'
```
ehlertjd: un = `username`? slightly more descriptive variable name would help

api/jobs/gears.py
```diff
@@ -222,3 +224,58 @@ def check_for_gear_insertion(doc):
...
+    blob_id = str(dxf.get_digest(pointer))
+    if dxf.blob_size(blob_id) > (10 * 1000* 1000): # 10 MB to bytes
+        raise Exception('Manifest is larger than 10 MB. Possible registry error?')
+
+    # Pull and assemble the manifest
+    raw_blob, _ = dxf.pull_blob(blob_id, size=True)
+    manifest = json.loads(''.join(raw_blob))
+
+    # Compatibility checks for the gears platform
+    assert manifest['architecture'] == 'amd64'
```
ehlertjd: I would prefer check, log and raise to assert.

api/jobs/handlers.py
```diff
@@ -73,6 +74,105 @@ def check(self):
...
+        """
+
+        # Allow for just a summary of gear names
+        gear_names_only = self.is_true('gear_names_only')
+
+        result = config.db.gear_tickets.find({
+            'origin': self.origin
+        })
+
+        if gear_names_only:
```
ehlertjd: Mutating return type based on query parameter can be problematic for swagger code generation. Is there an alternative (e.g. different endpoint?)

api/jobs/handlers.py
```diff
@@ -73,6 +74,105 @@ def check(self):
...
+        pointer = submit['pointer']
+
+        try:
+            _, image = confirm_registry_asset(repo, pointer)
+        except Exception as err:
+            raise InputValidationException(cause=err)
+
+        # import json
+        # self.log.debug(json.dumps(manifest, indent=4, sort_keys=True))
+        # self.log.debug(json.dumps(image, indent=4, sort_keys=True))
```
ehlertjd: cleanup commented code

api/jobs/handlers.py
```diff
@@ -73,6 +74,105 @@ def check(self):
...
+
+        geardoc = self.request.json
+
+        if geardoc.get('category') is None:
+            geardoc['category'] = 'converter'
+
+        check_for_gear_insertion(geardoc)
+
+        ticket = config.db.gear_tickets.insert_one({
+            'origin': self.origin,
```
ehlertjd: Set `timestamp` field...

api/api.py
```diff
@@ -163,9 +163,20 @@ def prefix(path, routes):
...
         route('/gears',                                  GearsHandler),
-        route('/gears/check',                            GearsHandler, h='check', m=['POST']),
-        route('/gears/temp',                             GearHandler, h='upload', m=['POST']),
-        route('/gears/temp/<cid:{oid}>',  GearHandler, h='download', m=['GET']),
+
+        # New upload flow
+        route('/gears/prepare-add',                      GearsHandler, h='prepare_add',     m=['POST']),
+        route('/gears/ticket/<_id:[^/]+>',               GearsHandler, h='get_ticket',      m=['GET']),
+        route('/gears/my-tickets',                       GearsHandler, h='get_own_tickets', m=['GET']),
+        route('/gears/save',                             GearsHandler, h='save',            m=['POST']),
```
ehlertjd: Decided against the [alternate scheme](https://github.com/flywheel-io/core/pull/1418#discussion_r225307102)?

swagger/paths/gears.yaml
```diff
@@ -13,6 +13,58 @@
...
 
+/gears/prepare-add:
+  post:
+    summary: Prepare a gear upload
+    operationId: prepare_add_gear
+    tags:
+    - gears
+    responses:
+      '200':
+        description: 'Returns a gear ticket ID'
```
ehlertjd: Good start, however SDK Code generation is reliant on having body and response schemas.

swagger/paths/gears.yaml
```diff
@@ -13,6 +13,58 @@
...
+      '200':
+        description: 'Returns a gear ticket'
+
+/gears/my-tickets:
+  get:
+    summary: Retrieve all gear tickets for the current user
+    operationId: get_my_gear_tickets
+    parameters:
+      - name: gear_names_only
+        in: path
```
ehlertjd: `in: query`

api/api.py
kofalt: Yeah

api/jobs/handlers.py
kofalt: This is 90% likely to be turned on, on some prod site during a rollout... I just don't want to clutter the logs during normal operation. I can just turn them on for now, adding a gear is a rare task

api/jobs/handlers.py
kofalt: Good point, easiest change would be to just return a string array and not have the list endpoint for now.

api/jobs/gears.py
kofalt: Any complaint about assert in particular? There will probably be more comparisons down the road & writing larger conditionals feels a waste. The exception is caught upstack to be turned into an API error.

swagger/paths/gears.yaml
kofalt: We'll see if some more decls get swagger to pass.

kofalt:
Comments addressed, save assert & swagger failures

ehlertjd:
LGTM!

PR #1472, opened 2018-12-10
New id if subject with id already exists in upgrade 60

hkethi002:
### Review Checklist


nagem:
We briefly talked about just always setting a new `_id` but I think this is fine/objectively ends with the same result. LGTM.

PR #1475, opened 2018-12-11
Billing report

ehlertjd:
## Deprecation
The old usage report endpoint has been moved to `/api/report/legacy-usage`

## Summary
In the new reports, storage usage is tracked as byte-days. (e.g. if a 1-byte file is stored on flywheel for three days, that would be 3 byte-days of storage, as would a 3-byte file stored for one day)
In addition, compute time and storage is divided into center and storage values. 
Center-pays compute is any job in the white list of gears below.
Center-pays storage is any file created by device, or as an output of one of the center-pays gears.
Everything else is treated as group-pays.

## Collection
Adds a new collection endpoint for usage report, intended to be called nightly by the cron container, at `/api/report/usage/collect`. Daily usage data is stored in the `usage_data` collection, indexed by year, month, group and project_id for fast aggregation. By default, the collect endpoint will collect yesterday's usage data. This can be overridden by supplying `year`, `month` and `day` query parameters. If an attempt is made to collect usage data where a record for that day already exists, then no record update will occur.
Two new metrics have been added to track usage collection:

- `fw_last_report_collection` indicates when the usage data was last collected successfully.
- `fw_report_collection_errors` is a counter that increases whenever a usage collection error occurs.

## Reporting
There are two new report endpoints. Both endpoints will report on a single month at a time, selected via `year` and `month` query parameters. Both endpoints support CSV download by passing the `csv=true` query parameter.
- `/api/report/usage` provides monthly summaries of usage by group and project. 
- `/api/report/daily-usage` provides detailed daily summaries of usage by group and project.

## Center-Pays Gear List
The center-pays gear list can be defined directly in the database by creating a `site` singleton, where `center_gears` is an array of algorithm names. If not set, then the default list will be `dcm2niix`, `dicom-mr-classifier` and `mriqc` for this PR.

An additional PR [has been](https://github.com/flywheel-io/core/pull/1476) drafted that supports setting these values via `/api/site/settings` endpoint.

## Utility Collection
A utility collection named `file_job_origin` has been created to facilitate joins between files and jobs, in order to determine which gear created the file. This collection is updated as part of the `collect` endpoint, prior to collecting that day's usage statistics.

### Review Checklist

### DIFF
--- a/api/reports/usage_report.py
+++ b/api/reports/usage_report.py
@@ -1,63 +1,85 @@
-import copy
 import datetime
-import dateutil
+
+from pymongo.operations import UpdateOne
 
 from .report import Report
-
 from .. import config
+from ..web.errors import APIReportParamsException
+from ..jobs import file_job_origin
 
-from ..web.errors import APIReportException, APIReportParamsException
+log = config.log
 
-
+BULK_UPDATE_BLOCK_SIZE = 100
 BYTES_IN_MEGABYTE = float(1<<20)
+MILLISECONDS_IN_HOUR = float(1000 * 60 * 60)
 
+# Tuple of collection name, file size key
+FILE_COLLECTIONS = ['projects', 'subjects', 'sessions', 'acquisitions', 'analyses']
 
 class UsageReport(Report):
     """
-    Creates a usage report, aggregated by month or project.
+    Creates a site usage report, aggregated by group and project.
 
-    Specify a date range to only return stats for that range.
+    Specify a year/month to get a report for that time period.
 
     Report includes:
-      - count of gears executed (jobs completed successfully)
-      - count of sessions
-      - aggregation of file size in megabytes
+        - Count of sessions
+        - Count of complete jobs, grouped by data/analysis
+        - Aggregation of gear execution time in hours, grouped by center/group 
+        - Aggregation of total storage size in megabytes, grouped by center/group 
+
+    Collection is done nightly, and stored in the usage_data database collection, using the below structure:
+        group: The group id
+        project: The project id
+        project_label: The project label (at last collection)
+        year: The collection year
+        month: The collection month
+        days: A map of day_of_month to collection record for that date, as follows:
+            session_count: The number of sessions that exist at time of collection
+            center_job_count: The number of jobs ran attributed to the center
+            group_job_count: The number of jobs ran attributed to the group
+            center_storage_bytes: The size of storage usage allocated to the center, in bytes
+            group_storage_bytes: The size of storage usage allocated to the group, in bytes
+            center_compute_ms: The amount of compute used attributed to the center, in seconds
+            group_compute_ms: The amount of compute used attributed to the group, in seconds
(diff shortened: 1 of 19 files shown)


api/reports/usage_report.py
```diff
@@ -67,300 +111,349 @@ def user_can_generate(self, uid):
...
+            group = row.pop('group')
+            project = row.pop('project')
+            project_label = row.pop('project_label')
+
+            query = {
+                'group': group,
+                'project': bson.ObjectId(project),
+                'year': year,
+                'month': month,
+                day_entry: {'$exists': False}
```
hkethi002: If we have this filter here, we may want to prevent users from giving dates to the collect operation that are not finished yet, i.e. if I run the collect during the middle of the day and somebody upload files afterwards, that won't be reflected until the next day.

nagem:
>If an attempt is made to collect usage data where a record for that day already exists, then no record update will occur.

Let's say there was a bug in the "collect" code. Would the fix be to delete the days with "bugged" information and rerun over that range?

ehlertjd:
@nagem Yes, that would work, with the caveat that the closer to the date the data is collected the more accurate it is. (i.e. if files move from one project to another that would be reflected in the re-collected data)

api/reports/usage_report.py
```diff
@@ -67,300 +111,349 @@ def user_can_generate(self, uid):
...
-            if 'created' in base_query:
-                file_q['files.created'] = base_query['created']
-                analysis_q['analyses.created'] = base_query['created']
+    @staticmethod
+    def _get_file_size_counts(coll_name, end_date):
+        """
+        Get total file size in bytes, for files created before end_date
+        Grouped by project_id, origin, gear_name and gear_version
+        """
+        file_q = {'files.created': {'$lt': end_date}}
```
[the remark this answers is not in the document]
ehlertjd: @hkethi002 We're deliberately reporting on deleted files, since they're still consuming storage until purged.

ehlertjd:
@hkethi002 Comments have been addressed

ryansanford:
What is the boundary that determines a day? UTC midnight to midnight to help avoid overlap and gaps? 

If `/api/report/usage/collect` is not called for several days or weeks due to some issue, Is it completely up to the client to detect which days have missing reports, and execute it accordingly? Are the endpoints/methods available to facilitate that?

swagger/paths/report.yaml
```diff
@@ -89,12 +89,112 @@
...
+      - in: query
+        type: integer
+        name: day
+        description: The day portion of the date
+    produces:
+      - text/event-stream
+    responses:
+      '200':
+        description: ''
+
```
nagem: This is endpoint intended for UIs that would list reports that are available for viewing?

ehlertjd: Yeah, that's correct. At @jody-zeitler's suggestion, it's much better than forcing the user to guess whether or not data was collected for that month.

PR #1483, opened 2018-12-20
Update Travis Pipeline Logic

mjburling:
It looks as though `gem` options `--no-rdoc` and  `--no-ri` have been beyond deprecated but are also causing errors within the `before_install` step. While troubleshooting this `gem` issue, I've also updated the version to latest stable 1.8.9.

Branch name was from the original summation that we could just drop the travis gem. Not so!


mjburling:
You bet I'm gonna squash this ugly thing pending approval.

PR #1485, opened 2019-01-02
Training set from search endpoint

hkethi002:
Takes a query or set of nodes and uploads a file containing the training set of the nodes, either returned from the query or given in the request.

### Review Checklist


api/handlers/dataexplorerhandler.py
```diff
@@ -615,6 +618,77 @@ def search(self):
...
+                if field in value:
+                    value = value[field]
+                else:
+                    return None
+            return value
+
+        def format_file_doc(file_source, labels):
+            """Format the elastic file doc into only the fields we want and where we want them"""
+            return {
+                'type': file_source['parent']['type'],
```
ehlertjd: I wonder if this should be `parent_type` and `parent_id` - would make a potential transition to formalized files easier in the future.

PR #1486, opened 2019-01-02
Accomodate dynamic type from urlparse().port

kofalt:
When the return from `urlparse().port` would match what urlparse expects for `urlparse().scheme`, port will be of type `None` instead of type `int`. This gets stringified to `"None"` in some parts of our code, causing an error.

This change allows for `None` or `443`, essentially hiding the dynamic type behavior.


nagem:
LGTM

PR #1488, opened 2019-01-03
Session creation on deleted subjects

hkethi002:
Packfile upload was not handling creation of sessions on deleted subject correctly

### Coming soon
A matrix of session and subject against operations and the expected outcomes

### Review Checklist


api/dao/containerstorage.py
```diff
@@ -146,12 +146,20 @@ def _from_mongo(self, cont):
                 cont['label'] = 'unknown'
 
     def create_or_update_el(self, payload, **kwargs):
-        if self.dbc.find_one({'_id': payload['_id']}):
+        if self.dbc.find_one({'_id': payload['_id'], 'deleted': {'$exists': False}}):
             payload['modified'] = datetime.datetime.utcnow()
             # Pop _id from mongo payload (immutable - would raise error)
             payload_copy = copy.deepcopy(payload)
             _id = payload_copy.pop('_id')
             return super(SubjectStorage, self).update_el(_id, payload_copy, **kwargs)
```
nagem: Is there ever a situation where the payload would have a pre-determined `_id`? I'm not sure this needs to be a different `if` statement than the one below it.

tests/integration_tests/python/test_uploads.py
```diff
@@ -1464,6 +1465,12 @@ def test_packfile_upload(data_builder, file_form, as_admin, as_root, api_db):
     project = data_builder.create_project(group=group)
     session = data_builder.create_session()
 
+    subject = data_builder.create_subject(project=project, code='subj-01')
+    r = as_admin.delete('/subjects/' + subject)
+    assert r.ok
+
+    number_of_sessions = len(list(api_db.subjects.find({'project': bson.ObjectId(project)})))
```
nagem: this should be `number_of_subjects`

nagem:
In the initial review I had a few questions/comments. I think the approach to create a new subject if an old one exists and is deleted is correct.

api/dao/containerstorage.py
hkethi002: When it goes through the extract_subject logic in containerutil, it predetermines the id and sets session.subject before creating the subject container.

api/dao/containerstorage.py
```diff
@@ -146,12 +146,20 @@ def _from_mongo(self, cont):
...
+        elif self.dbc.find_one({'_id': payload['_id']}):
+            # If the subject exists but is deleted, create a new subject with a new id
+            # This logic should only be used by the placers/hierarchy.py
+            # Other subject operations should be going through
+            # extract_subject which prevents this from happening
+            payload.pop('_id')
+            payload['created'] = payload['modified'] = datetime.datetime.utcnow()
```
ehlertjd: could remove duplicate lines here by just popping the `_id` then dropping into the `else` case.

api/dao/containerutil.py
```diff
@@ -119,11 +119,19 @@ def extract_subject(session, project):
...
+        if result:
+            subject['_id'] = bson.ObjectId()
     elif subject.get('code'):
-        query = {'code': subject['code'], 'project': project['_id']}
+        # If a non-deleted subject with that code doesn't exist in the project, create a new id
+        query = {'code': subject['code'], 'project': project['_id'], 'deleted': {'$exists': False}}
         result = config.db.subjects.find_one(query)
         if result:
             subject['_id'] = result['_id']
+        else:
```
ehlertjd: These 2 lines are effectively redundant with 135-136, aren't they?

PR #1489, opened 2019-01-03
[FLYW-142] Add flywheel roles

hkethi002:
Instead of a root boolean, there are roles:

| Root Bool | Role Enum |
|---|---|
| root=false | role=user |
|  root=true | role=site-admin  |
|  new Role| role=developer  |

The Developer role is identical to the user role except that it can also create gears

### ToDo
- [x] Fix sdk tests

### Review Checklist

### DIFF
--- a/api/auth/__init__.py
+++ b/api/auth/__init__.py
@@ -1,5 +1,29 @@
+from enum import Enum
+
 from ..types import Origin
 from ..web.errors import APIPermissionException
+from ..config import log
+
+# The values for the role privileges are below
+class Privilege(Enum):
+
+    is_user =           0b0001
+    can_upload_gear =   0b0010
+    is_admin =          0b0100
+    is_drone =          0b1000
+
+# the values for the user roles and what they can access are below
+class Role(Enum):
+    public =       ('public', 0)
+    user =         ('user', Privilege.is_user.value)
+    developer =    ('developer', Privilege.is_user.value | Privilege.can_upload_gear.value)
+    site_admin =   ('site_admin', Privilege.is_user.value | Privilege.can_upload_gear.value | Privilege.is_admin.value)
+    drone =        ('drone', Privilege.is_user.value | Privilege.can_upload_gear.value | Privilege.is_admin.value | Privilege.is_drone.value)
+
+    def __init__(self, role, privileges):
+        self.role = role
+        self.privileges = privileges
+
 
 PERMISSIONS = [
     {
@@ -47,40 +71,22 @@ def always_ok(exec_op):
     """
     return exec_op
 
-def require_login(handler_method):
+def require_privilege(privilege):
     """
-    A decorator to ensure the request is not a public request.
+    Returns a decorator that requires the given role
+    """
+    def require_privilege_decorator(handler_method):
+        def check_role(self, *args, **kwargs):
+            if not any([role.privileges & privilege.value == privilege.value for role in self.roles]):
+                raise APIPermissionException('{} privilege required for action.'.format(privilege.name))
+            return handler_method(self, *args, **kwargs)
+        return check_role
+    return require_privilege_decorator
 
-    Accepts drone and user requests.
+def has_privilege(roles, privilege):
     """
-    def check_login(self, *args, **kwargs):
-        if self.public_request:
-            raise APIPermissionException('Login required.')
-        return handler_method(self, *args, **kwargs)
-    return check_login
-
-def require_admin(handler_method):
+    A non decorator role check function
     """
-    A decorator to ensure the request is made as a site admin.
-
-    Accepts drone and user requests.
-    """
-    def check_admin(self, *args, **kwargs):
-        if not self.user_is_admin:
-            raise APIPermissionException('Admin user required.')
-        return handler_method(self, *args, **kwargs)
-    return check_admin
-
-def require_drone(handler_method):
-    """
-    A decorator to ensure the request is made as a drone.
-
-    Will also ensure site admin, which is implied with a drone request.
-    """
-    def check_drone(self, *args, **kwargs):
-        if self.origin.get('type', '') != Origin.device:
-            raise APIPermissionException('Drone request required.')
-        if not self.user_is_admin:
-            raise APIPermissionException('Superuser required.')
-        return handler_method(self, *args, **kwargs)
-    return check_drone
+    if not any([role.privileges & privilege.value == privilege.value for role in roles]):
+        raise APIPermissionException('{} privilege required for action.'.format(privilege))
+    return True
--- a/api/handlers/userhandler.py
+++ b/api/handlers/userhandler.py
@@ -7,7 +7,7 @@ from ..web import base
 from .. import util
 from .. import config
 from .. import validators
-from ..auth import userauth, require_admin, require_login
+from ..auth import userauth, require_privilege, Privilege
 from ..auth.apikeys import UserApiKey
 from ..auth.authproviders import AuthProvider
 from ..dao import containerstorage
@@ -37,6 +37,7 @@ class UserHandler(base.RequestHandler):
         result = permchecker(self.storage.exec_op)('GET', _id, projection=projection or None)
         if result is None:
             self.abort(404, 'User does not exist')
+        result['root'] = 'site_admin' in result.get('roles', ['user'])
         return result
 
     def self(self):
@@ -46,6 +47,7 @@ class UserHandler(base.RequestHandler):
         user = self.storage.exec_op('GET', self.uid)
         if not user:
             self.abort(403, 'user does not exist')
+        user['root'] = 'site_admin' in user.get('roles', ['user'])
         api_key = UserApiKey.get(self.uid)
         if api_key:
             user['api_key'] = {
@@ -61,6 +63,8 @@ class UserHandler(base.RequestHandler):
         if not self.user_is_admin:
             projection['wechat'] = 0
         page = permchecker(self.storage.exec_op)('GET', projection=projection, pagination=self.pagination)
+        for result in page['results']:
+            result['root'] = 'site_admin' in result.get('roles', ['user'])
         if page['results'] is None:
             self.abort(404, 'Not found')
         return self.format_page(page)
@@ -89,6 +93,8 @@ class UserHandler(base.RequestHandler):
         # Password is only valid if basic auth enabled
         password = payload.pop('password', None)
 
+        if payload.get('root') is not None:
+            payload['roles'] = ['site_admin'] if payload.pop('root', None) else ['user']
         mongo_schema_uri = validators.schema_uri('mongo', 'user.json')
         mongo_validator = validators.decorator_from_schema_path(mongo_schema_uri)
         payload_schema_uri = validators.schema_uri('input', 'user-update.json')
@@ -113,6 +119,9 @@ class UserHandler(base.RequestHandler):
 
         # Password is only valid if basic auth enabled
         password = payload.pop('password', None)
+        payload['roles'] = payload.get('roles', ['user'])
+        if payload.pop('root', None):
+            payload['roles'] = ['site_admin']
 
         if self.is_true('wechat'):
             payload['wechat'] = {'registration_code': base64.urlsafe_b64encode(os.urandom(42))}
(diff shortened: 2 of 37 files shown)


api/auth/__init__.py
```diff
@@ -60,6 +60,18 @@ def check_login(self, *args, **kwargs):
...
+    a decorator to ensure the request is made as a developer.
+
+    Accepts drone, site-admin, and developer requests.
+    """
+    def check_developer(self, *args, **kwargs):
+        if not self.user_is_developer:
+            raise APIPermissionException('Developer role required.')
+        return handler_method(self, *args, **kwargs)
+    return check_developer
+
```
kofalt: Suggest a role enum, and replacing `require_developer` / `require_admin` / et all with a single `require_role`.

api/handlers/reporthandler.py
```diff
@@ -163,7 +163,7 @@ def user_can_generate(self, uid):
         """
         User generating report must be superuser
         """
-        if config.db.users.count({'_id': uid, 'root': True}) > 0:
+        if config.db.users.count({'_id': uid, 'roles': 'site-admin'}) > 0:
```
kofalt: If an enum is adopted as above, use it here & everywhere we have a role string

jody-zeitler:
I don't think it's necessary for this to be a completely breaking change. Could the migration logic be applied to the User endpoints to retain support for the `root` boolean property -- convert `root: true` to the `site-admin` role for PUTs and derive it for GETs -- and just mark it as deprecated until the next major release?

kofalt:
I would also like pretty much everything in Justin's ^ comment to end up in a giant comment block somewhere please :)

jody-zeitler:
`/api/users/self` will need the derived `root` property as well.

api/auth/__init__.py
```diff
@@ -1,5 +1,27 @@
...
+    super_user =    0b01111
+    drone =         0b11111
+
+# The values for the role privileges are below
+class Privilege(Enum):
+    user =         0b00001
+    developer =    0b00010
+    site_admin =   0b00100
+    super_user =   0b01000
+    drone =        0b10000
```
roycollins: One suggestion I would have is abstract the permissions from the roles entirely.  The permissions could be something like write_gears, read_gears, create_reports, view_reports.  You could even have a general read_data, write_data that handlers use until you have the need for more granular permissions.   The permission then would be a hashmap lookup of permission  based on the role.

This give the ability to customize permissions in the near future. But more so it allows you to customize the role with a simple key in a hash map.  Say if a customer wants a role that allows a user to view subject data and be able to upload attachments but not able to upload dicom data.  Or any combination they can come up with.  You could even allow custom role definitions with a check box that maps the keys to the role.  Then again on permission check its again an O(1) average lookup to see if the permission is valid for that user.  

You could add role inheritance too if you wanted but I would save that for version 2.  

Just my thoughts on making the permissions more scalable as we grow without adding a lot of extra overhead.

api/auth/__init__.py
```diff
@@ -47,53 +69,24 @@ def always_ok(exec_op):
...
-            raise APIPermissionException('Login required.')
-        return handler_method(self, *args, **kwargs)
-    return check_login
-
-def require_admin(handler_method):
-    """
-    A decorator to ensure the request is made as superuser.
-
-    Accepts drone and user requests.
+def require_privilege_decorator(privilege):
```
ehlertjd: Since this is very common, let's drop the `decorator` suffix and save some typing.

There's also a an opportunity to unit test these functions, I'd love to see us start doing that more often.

api/auth/__init__.py
ehlertjd: I think in a lot of ways we're getting closer to on-track, however I'm largely in agreement with Roy. In the above, the permissions and roles are pretty much identical, when I think we truly want `Role` to be a composition of Privileges/Permissions.

We can keep Privileges simple to start, and add increased granularity when it is appropriate.
I would consider the following permissions for today:
- **is_user**: Current concept of has a user account (is logged in)
- **is_admin**: Whether or not this is a site admin account. Do we really need a distinction between this and `super_user`?
- **can_upload_gears** - This is what we're trying to add today. Developer and Admin roles would have this permission.

If you keep roles as an Enum, then I would set the permissions as an extra attribute, since there's no reason that those bitmasks would have to be unique. I would also express a role as an OR'ed list of bitmasks (see [Enum docs](https://docs.python.org/3/library/enum.html#planet).) e.g.
```python
class Role(Enum):
  user = ('user', Privilege.is_user)
  developer = ('developer', Privilege.is_user | Privilege.upload_gear)
  admin = ('admin', Privilege.is_user | Privilege.upload_gear | Privilege.is_admin)

  def __init__(self, name, permissions):
    self.name = name
    self.permissions = permissions
```

api/auth/__init__.py
```diff
@@ -47,53 +69,24 @@ def always_ok(exec_op):
...
-        return handler_method(self, *args, **kwargs)
-    return check_admin
-
-def require_superuser(handler_method):
+    Returns a decorator that requires the given role
     """
-    A decorator to ensure the request is made as superuser.
+    def require_privilege(handler_method):
+        def check_role(self, *args, **kwargs):
+            log.debug(self.role.value)
```
ehlertjd: Remove log statement or add context.

api/auth/__init__.py
```diff
@@ -47,53 +69,24 @@ def always_ok(exec_op):
...
+            if self.role.value & privilege.value != privilege.value:
+                raise APIPermissionException('{} privilege required for action.'.format(privilege))
+            return handler_method(self, *args, **kwargs)
+        return check_role
+    return require_privilege
 
-    Accepts drone and user requests.
+def require_privilege_check(role, privilege):
```
ehlertjd: Maybe just `has_privilege`?

api/auth/__init__.py
```diff
@@ -1,5 +1,27 @@
...
+class Role(Enum):
+    public =        0b00000
+    user =          0b00001
+    developer =     0b00011
+    site_admin =    0b00111
+    super_user =    0b01111
+    drone =         0b11111
+
+# The values for the role privileges are below
+class Privilege(Enum):
```
ehlertjd: Let's add a comment for each privilege describing what it permits you to do in terms of functionality.

api/auth/__init__.py
kofalt: General agreement with everything in this thread :+1: I like the abstraction between permissions / roles.

nagem: >* **is_admin**: Whether or not this is a site admin account. Do we really need a distinction between this and `super_user`?

No, IMO. That is legacy code that was written to determine the difference between devices (that were auto-granted superuser) and users that were admin. Let's get rid of the superuser distinction.

hkethi002:
CodeCov not behaving correctly, please ignore it and this PR is ready for review again.

api/auth/__init__.py
```diff
@@ -47,53 +71,24 @@ def always_ok(exec_op):
...
-    def check_login(self, *args, **kwargs):
-        if self.public_request:
-            raise APIPermissionException('Login required.')
-        return handler_method(self, *args, **kwargs)
-    return check_login
+    def require_privilege(handler_method):
+        def check_role(self, *args, **kwargs):
+            log.debug(self.role.privileges)
+            if self.role.privileges & privilege.value != privilege.value:
+                raise APIPermissionException('{} privilege required for action.'.format(privilege))
```
ehlertjd: `privilege.name`

api/handlers/userhandler.py
```diff
@@ -99,6 +103,10 @@ def post(self):
         """Add user"""
         permchecker = userauth.default(self)
         payload = self.request.json_body
+        payload['role'] = payload.get('role', 'user')
+        if payload.get('root'):
```
ehlertjd: Should we `pop('root')` here?

api/web/base.py
```diff
@@ -116,31 +116,35 @@ def initialization_auth(self):
...
             self.user_is_admin = True
+            self.role = Role.drone
         elif self.scope is not None:
             self.superuser_request = False
             self.user_is_admin = False
+            self.role = Role.user
         else:
-            user = config.db.users.find_one({'_id': self.uid}, ['root', 'disabled'])
+            user = config.db.users.find_one({'_id': self.uid}, ['role', 'disabled'])
+            self.log.debug('RUser request has privkledr orle of {}'.format(user.get('role')))
```
ehlertjd: Do what now?

jody-zeitler:
@alonna-frazier has been periodically testing this with the UI branch on test1. This is ready from a frontend perspective.

api/jobs/handlers.py
```diff
@@ -594,21 +594,51 @@ def ask(self):
...
+    @require_privilege(Privilege.is_admin)
     def reap_stale(self):
         count = Queue.scan_for_orphans()
         return { 'orphaned': count }
 
 class JobHandler(base.RequestHandler):
     """Provides /Jobs/<jid> routes."""
 
-    @require_admin
+    def _log_job_access(self, job):
```
ehlertjd: Should this function be here? Is it used?

api/web/base.py
```diff
@@ -116,33 +116,38 @@ def initialization_auth(self):
...
             self.complete_list = True
+            self.roles.append(Role.drone)
         elif self.scope is not None:
             self.user_is_admin = False
             self.complete_list = False
+            self.roles.append(Role.user)
         else:
-            user = config.db.users.find_one({'_id': self.uid}, ['root', 'disabled'])
+            user = config.db.users.find_one({'_id': self.uid}, ['roles', 'disabled'])
+            self.log.debug('User\'s roles are {}'.format(user.get('roles')))
```
ehlertjd: I think we want to remove this debug statement, right?

api/web/base.py
```diff
@@ -116,33 +116,38 @@ def initialization_auth(self):
...
                 self.user_is_admin = False
-            if self.is_true('root') or self.is_true('exhaustive'):
-                if self.user_is_admin:
-                    self.complete_list = True
-                else:
-                    self.abort(403, 'user ' + self.uid + ' is not authorized to request complete lists.')
+
+            self.roles += [Role[role] for role in user.get('roles', ['user'])]
+
+            if (self.is_true('root') or self.is_true('exhaustive')) and has_privilege(self.roles, Privilege.is_admin):
```
ehlertjd: Should we keep the 403 for unauthorized request of complete lists?

api/web/base.py
```diff
@@ -116,33 +116,38 @@ def initialization_auth(self):
...
+                self.complete_list = True
             else:
                 self.complete_list = False
+            self.log.debug('Returning exhaustive lists: {}'.format(self.complete_list))
```
ehlertjd: Remove debug statement.

bin/database.py
```diff
@@ -2330,6 +2330,18 @@ def upgrade_to_63():
     cursor = config.db.projects.find({'template': {'$exists': True}})
     process_cursor(cursor, upgrade_template_to_list)
 
+def upgrade_to_64():
+    """Update all users to having a role, non root -> user and root -> site-admin"""
+    """Assumptions
+        - user.root is boolean if it exists
+    A failure of these assumption won't result in the DB upgrade failing,
+    but the db will be in an inconsistent state
```
ehlertjd: 👌

api/web/base.py
hkethi002: I think so, it should be doing that still, I'll add some tests for it though

api/jobs/handlers.py
hkethi002: Yeah, it gets used [here](https://github.com/flywheel-io/core/pull/1489/files/12bc61ed857c9f2861a667e4c6a5475c63c120c3#diff-8ae88fa922271b27b6066aae46fda7abR638)

hkethi002: I think I made a mistake in the rebase

hkethi002:
Nope, the api should accept `root: true` and convert that to a `site-admin` role

alonna-frazier:
Went ahead and tested the frontend on test1.dev. Roles appear to behave as expected after a couple smoke testing passes. I've noted QA approval on the frontend PR.

PR #1491, opened 2019-01-07
Disable codecov so that builds can actually succeed

kofalt:
Ya'll are wild, you know that?

wild

image

image


ehlertjd:
Looks like it might be the [concurrency setting](https://coverage.readthedocs.io/en/coverage-4.2/config.html#run)
Verifying, then I'll open a PR.

kofalt:
With #1493 I think this PR's time has come ;(

nightmare

PR #1492, opened 2019-01-07
Add detail endpoint for jobs.

ehlertjd:
This adds /api/jobs/{JobId}/detail, which will provide additional
information on a job, including retrieving the current parent labels
and file objects for inputs and outputs.

### Review Checklist


api/jobs/handlers.py
```diff
@@ -577,6 +578,122 @@ class JobHandler(base.RequestHandler):
...
+                    continue
+
+                # Retrieve the container
+                cont = get_container(ref)
+                if cont is None:
+                    continue
+
+                # Check access, (Raises APIPermissionException)
+                if not authorized and ref.check_access(self.uid, 'ro', cont=cont):
+                    authorized = True
```
nagem: So a user is authorized if they have access to one of the inputs of a job? I'm not too worried about cross-project jobs being issues here but wanted to confirm your thinking.

api/jobs/handlers.py
```diff
@@ -577,6 +578,122 @@ class JobHandler(base.RequestHandler):
...
+
+                if hasattr(ref, 'get_file'):
+                    # Raises APINotFoundException
+                    try:
+                        rec['object'] = ref.get_file(container=cont)
+                        rec['object'].pop('info', None)  # Remove info, if present
```
nagem: Is the info pop to prevent having to log_access?

api/jobs/handlers.py
ehlertjd: @nagem I see your point, this would check if *any* input, whereas [_log_read_check](https://github.com/flywheel-io/core/blob/master/api/jobs/handlers.py#L683-L695) checks *all*. I'll change that.

api/jobs/handlers.py
ehlertjd: Yes, that was the desire.

ehlertjd:
@nagem Updated to authorize for any retrieved container (if not admin)

api/jobs/handlers.py
```diff
@@ -577,6 +578,123 @@ class JobHandler(base.RequestHandler):
...
+                        log.debug('Unable to retrieve file on container: type=%s, id=%s, name=%s',
+                            ref.type, ref.id, ref.name)
+                else:
+                    rec['object'] = cont
+
+        # If we're still not authorized, check the destination
+        dest_cont = get_container(job.destination)
+        if dest_cont and not self.user_is_admin:
+            # Raises APIPermissionException
+            job.destination.check_access(self.uid, 'ro', cont=dest_cont)
```
hkethi002: Is it possible for authorized to be set to true and the check_access call raising the Exception, or should `and not authorized` be added to the if statement?

ehlertjd: The `authorized` variable is tracking whether *any* authorization has occurred. So the logic is:
- If the user is admin, they're authorized.
- If the user is authorized for **all** retrievable containers, they're authorized.
- If the user is not authorized for **all** containers, or we couldn't retrieve any containers, then they're not authorized.

api/jobs/handlers.py
```diff
@@ -577,6 +578,123 @@ class JobHandler(base.RequestHandler):
...
+        elif not authorized:
+            # Couldn't find destination container, and cannot check access
+            raise APIPermissionException('User {} does not have access to job {}'.format(self.uid, _id))
```
nagem: Does `get_container` find deleted containers? Is a 403 on a job detail that is related to a deleted container expected?

ehlertjd: We would get a `403` if we couldn't retrieve *any* of the related containers, however I'm not sure how I feel about authorizing based on what a deleted container says.

PR #1493, opened 2019-01-07
Set concurrency type for coverage

ehlertjd:
Supersedes #1491. Turns out that we need to [set the concurrency type](https://coverage.readthedocs.io/en/coverage-4.2/config.html#run) for coverage, after we enabled gevents in gunicorn.

Now with **50% less snark!**

### Review Checklist


nagem:
Would we expect this codecov build to succeed in this PR?

kofalt:
We would ideally also expect <= 1 codecov comment AND a better meme than my PR _hands on hips_

ehlertjd:
I would expect the codecov to fail as this report is compared to whatever garbage is in master, and succeed in subsequent PRs into this commit. I would also expect that the coverage report be consistent between builds, which it appears to be by looking at a diff of the outputs, and by comparing local runs.

I'm not sure how to limit codecov to 1 comment. The [condition for submitting the report](https://github.com/flywheel-io/core/blob/master/.travis.yml#L55-L57) appears to be correct to me.

kofalt:
Hmm, maybe we should give them an email.

PR #1500, opened 2019-01-11
[FLYW-200] Fix DB Upgrade 62

ehlertjd:
Executes `fix_parents` before running the upgrade, and be tolerant of orphan containers and collections.

### Review Checklist


bin/database.py
```diff
@@ -2328,7 +2328,7 @@ def set_job_containers_62(cont, cont_name):
         if not c:
             continue
 
-        for parent_id in c['parents'].itervalues():
+        for parent_id in c.get('parents', {}).itervalues():
```
mjburling: 👍 I imagine without a schema, accessing by `['keyword']` is pretty unsafe.

PR #1497, opened 2019-01-10
Make elapsed optional on job tickets

ehlertjd:
This makes the `elapsed` time optional on job completion tickets in a backward compatible manner. 
If `elapsed` is specified and no `profile` is given, then `elapsed will be used.
Otherwise `profile.elapsed_time_ms` will take precedence.

### Review Checklist

- Tests were added to cover all code changes
- Documentation was added / updated
- Code and tests follow standards in CONTRIBUTING.md
- Extra care is taken and log messages are created for [Critical Data](https://github.com/flywheel-io/core/wiki/Critical-Data-Flows)

PR #1503, opened 2019-01-12
[FLYW-207] Job Completion Tickets

ehlertjd:
Supersedes #1497. 

This updates the job ticketing workflow as follows:
- `/prepare-complete` no longer looks at success, elapsed or profile information in input.
- `/engine` does not update the state of a job (although it does automatically update some profile info)
- `/complete` finalizes the job, transitioning to a final state and updating profile information.

Uploads to placers will no longer cause `500` errors if there are invalid rules on the process.

Details about files that were saved, and rules that failed to evaluate are now added to the job logs.

While this is a breaking change in the API, in practice the prior job ticketing workflow was not being used. This API is backward compatible with the legacy behavior.

This feature is advertised as available in the site config features list as `job_tickets`.

### Review Checklist


api/api.py
```diff
@@ -145,24 +145,25 @@ def prefix(path, routes):
...
+            route('/<:[^/]+>',                      JobHandler),
+            route('/<:[^/]+>/config.json',          JobHandler,  h='get_config'),
+            route('/<:[^/]+>/retry',                JobHandler,  h='retry',                m=['POST']),
+            route('/<:[^/]+>/logs',                 JobHandler,  h='get_logs',             m=['GET']),
+            route('/<:[^/]+>/logs/text',            JobHandler,  h='get_logs_text',        m=['GET']),
+            route('/<:[^/]+>/logs/html',            JobHandler,  h='get_logs_html',        m=['GET']),
+            route('/<:[^/]+>/logs',                 JobHandler,  h='add_logs',             m=['POST']),
             route('/<:[^/]+>/prepare-complete',     JobHandler,  h='prepare_complete',     m=['POST']),
+            route('/<:[^/]+>/complete',             JobHandler,  h='complete',             m=['POST']),
             route('/<:[^/]+>/accept-failed-output', JobHandler,  h='accept_failed_output', m=['POST']),
```
nagem: You'll want to remove the `accept-failed-output` route up here as well.

ehlertjd: I did re-implement it to return `410` I don't know how nice we wanted to be about it. I'd be just as happy removing it.

nagem: I commented too soon before seeing the reimplementation below the removal. The `410` is a good practice, we can leave it and remove it in the future.

kofalt: I'd just be harsher and remove it, since no callee actually used it. But IDC

api/config.py
```diff
@@ -252,12 +252,20 @@ def get_public_config():
...
+        'signed_url': hasattr(fs, 'get_signed_url'),
+    }
+
     return {
         'created': __config.get('created'),
         'modified': __config.get('modified'),
         'site': __config.get('site'),
         'auth': auth,
-        'signed_url': hasattr(fs, 'get_signed_url'),
+        'signed_url': features['signed_url'],  # Legacy note: clients expect top-level signed_url key
```
kofalt: FWIW, I'll just rig for the `features` map, but I assume by "legacy" you mean "some versions of the web frontend / sdks" ?

ehlertjd: @kofalt Yep, namely reapers and CLI rely on the top-level `signed_url`

kofalt: Alright. Filing a CLI PR to fix the later.

kofalt:
We've validated old/new component combinations, and are now making some final changes.

PR #1508, opened 2019-01-22
[FLYW-165] Add check status endpoint

hkethi002:
Use devices to track mongoconnector status

Endpoint returns a state of `'missing'`, `'indexing'` or `'up-to-date'`.

### Review Checklist


api/api.py
```diff
@@ -113,6 +113,7 @@ def prefix(path, routes):
         # Search
         route('/dataexplorer/search',                   DataExplorerHandler,   h='search',                 m=['POST']),
         route('/dataexplorer/facets',                   DataExplorerHandler,   h='get_facets',             m=['POST']),
+        route('/dataexplorer/search/status',            DataExplorerHandler,   h='check_search_status',    m=['GET']),
```
ehlertjd: Add swagger docs

api/handlers/dataexplorerhandler.py
```diff
@@ -418,6 +418,16 @@ def _parse_request(self, request_type='search', search_request=None):
 
         return return_type, modified_filters, search_string, size
 
+    def check_search_status(self):
```
ehlertjd: Probably would call it `get_search_status` to be consistent with HTTP verbs

api/handlers/dataexplorerhandler.py
```diff
@@ -418,6 +418,16 @@ def _parse_request(self, request_type='search', search_request=None):
...
+        storage = ContainerStorage('devices', use_object_id=True)
+        mongoconnector = storage.get_all_el({'name': 'mongo-connector'}, None, None)
+        if not mongoconnector:
+            raise APIPeripheralException('Mongo Connector has not registered itself with Core')
```
ehlertjd: I would just return a `200` with a different `'status'`. Returning `503` should be reserved to indicate that the API server is unavailable. Using it here could be a false-positive for some future API server status monitoring system.

hkethi002:
I would eventually like that to be the case, but I felt adding in the mongo-connector repo to this ticket would increase the scope, as the changes would have to be made there.

> What do we do when mongo connector exists with a non-zero status? Should we be publishing that an error occurred to core? Was that in the scope of this ticket?

I don't think so, the scope was just for a warning if mongo connector was still indexing. I think any type of non-zero status should probably be in the domain of prometheus.

PR #1510, opened 2019-01-23
[FLYW-327] Fix db upgrade 62

ehlertjd:
Very old jobs could have a dictionary for inputs, rather than a list.

### Review Checklist


ehlertjd:
@hkethi002 Please re-check; added additional upgrade logic for inputs

ehlertjd:
@hkethi002 Please re-check; added additional upgrade logic for inputs

bin/database.py
```diff
@@ -2306,8 +2306,27 @@ def set_job_containers_62(cont, cont_name):
...
+    if not isinstance(safe_inputs, list):
+        if safe_inputs:
+            logging.error('Encountered a non-empty, non-list set of inputs on job: %s', cont['_id'])
+
+        # Convert inputs to an array. See upgrade_to_11
+        if isinstance(safe_inputs, dict):
+            safe_inputs = []
+            for key, inp in cont['inputs'].iteritems():
+                inp['input'] = key
+                safe_inputs.append(inp)
```
hkethi002: Have we seen any examples of non-empty dictionary inputs, and if so, do we know if they have the other fields correct as well, such as 'type' and 'id' instead of 'container_type' and 'container_id'. I think these keys were renamed at some point as well.

kofalt: Checking vistalabs with the validator, I did not see non-empty `inputs` maps in the first 10 of 13 detected failures. Does not say anything about other customers or the last 3 of those failures.

kofalt:
Waiting on the sanity validator to run on each production site; if that succeeds I'll be checking off this PR.

kofalt:
Priority changed; the mongo-upgrade check seems to work on the originally failing customer; let's go for launch.

PR #1513, opened 2019-01-28
[FLYW-270] Rules multi input gears allowed if a fixed inputs list is provided

hkethi002:
### Changes
- Rules have a new field for fixed inputs, it is an array of job inputs. The type should be the container type of the file's parent container.
 ```
"fixed_inputs": [
  {
    "type": "project",
    "id": "0000000000000000000000000",
    "name": "fixed_input.txt",
    "input": "file"
  }
]
```
- For a rule to be valid, the number of fixed inputs **must** be one less than the number of file inputs the gear has. 
- All optional inputs are still ignored.
- Fixed inputs is mutually exclusive with auto-update, it's the same reason as configs on rules.

### Review Checklist

### CODE REVIEW

File: api/jobs/rules.py [hunk 1]
```diff
@@ -365,3 +374,19 @@ def validate_auto_update(rule_config, gear_id, update_gear_is_latest, current_ge
...
+def validate_fixed_inputs(geardoc, fixed_inputs):
+    """
+    Validates the fixed inputs for a rule given a gear doc
+    The fixed inputs must:
+        - add up to one less than the number of required gear inputs
+        - all be valid inputs for the gear
+    """
+    fixed_inputs = fixed_inputs if fixed_inputs else []
+    if gears.count_file_inputs(gears.filter_optional_inputs(geardoc)) - len(fixed_inputs) != 1:
+        raise InputValidationException("Rule must only have 1 non-fixed gear value")
```
ehlertjd: Rephrase? `Rule must have exactly 1 non-fixed gear value`

api/jobs/rules.py
```diff
@@ -355,8 +364,8 @@ def validate_regexes(rule):
         )
 
 
-def validate_auto_update(rule_config, gear_id, update_gear_is_latest, current_gear_is_latest):
-    if rule_config:
+def validate_auto_update(rule_config, gear_id, update_gear_is_latest, current_gear_is_latest, fixed_inputs):
+    if rule_config or fixed_inputs:
         raise InputValidationException("Gear rule cannot be auto-updated with a config")
```
ehlertjd: I would make this two separate checks and return a different message for each.

api/jobs/rules.py
```diff
@@ -365,3 +374,19 @@ def validate_auto_update(rule_config, gear_id, update_gear_is_latest, current_ge
...
         raise InputValidationException("Gear rule cannot be auto-updated unless it is uses the latest version of the gear")
+
+
+def validate_fixed_inputs(geardoc, fixed_inputs):
+    """
+    Validates the fixed inputs for a rule given a gear doc
+    The fixed inputs must:
+        - add up to one less than the number of required gear inputs
+        - all be valid inputs for the gear
+    """
```
ehlertjd: Should we add a check that they must refer to an existing file?

api/jobs/rules.py
```diff
@@ -149,28 +149,37 @@ def eval_rule(rule, file_, container):
...
+        if gear['gear']['inputs'][x]['base'] == 'file' and x not in fixed_input_keys:
             input_name = x
 
     inputs = {
         input_name: input_
     }
 
+    if fixed_inputs:
+        for fixed_input in fixed_inputs:
+            inputs[fixed_input['input']] = FileReference(type=fixed_input['type'], id=str(fixed_input['id']),
```
ehlertjd: I wonder if we should verify the existence of fixed inputs before queuing the job? If those ever get deleted, it'd be tough to figure out why jobs are failing.

jody-zeitler:
Will fixed inputs for batch runs be done in another PR?

ehlertjd:
@hkethi002 
1. I think that would be fine - it would require some coordination with the frontend to ensure that we're displaying a useful error message to the user in that situation.
2. Will the current gear suggest endpoint work for that - what should the suggested values be? What if you're creating a site-rule? Or can you even create a site rule with fixed input - are there permission implications there?

hkethi002:
Good point, I'm not sure how a site rule with fixed inputs works or even if it makes sense. I think a site container is eventually going to be the solution, then we'll also unblock site level context, and if there's ever a site wide license for a gear, but for now I think disabling fixed inputs for site level rules may be the best option. 

Also, I did some thinking about disabling the rules and I think disallowing files to be deleted is actually the correct way to go, the reason being that rules require project `admin` permissions, but deleting only needs `rw`, so it doesn't make sense for a `rw` user to be able to modify rules indirectly by deleting the fixed inputs.

api/handlers/listhandler.py
```diff
@@ -641,6 +642,10 @@ def delete(self, cont_name, list_name, **kwargs):
             analysis_ids = [str(a['_id']) for a in analyses]
             raise APIPermissionException('Cannot delete file {} referenced by analyses {}'.format(filename, analysis_ids),
                 errors={'reason': 'analysis_conflict'})
+        rules = Rules()
+        if rules.find(single_doc=True, fixed_inputs={'$elemMatch': {'name': filename, 'id': _id}}):
+            raise APIPermissionException('File is used as fixed input in a rule, please delete the rule to delete the file')
```
ehlertjd: Log / indicate which rule(s) - Ideally project label(s) and rule name. Otherwise this would be a beast to track down.

api/jobs/jobs.py
```diff
@@ -86,11 +86,6 @@ def __init__(self, gear, inputs, destination=None, tags=None,
         if profile is None:
             profile = {}
 
-        if destination is None and inputs is not None:
```
ehlertjd: Does this make `destination` required? If so, can we add a check/raise?

api/jobs/rules.py
```diff
@@ -149,28 +149,37 @@ def eval_rule(rule, file_, container):
...
 
     gear = gears.get_gear(gear_id)
     gear = gears.filter_optional_inputs(gear)
+    fixed_input_keys = [fixed_input['input'] for fixed_input in fixed_inputs] if fixed_inputs else []
 
-    if gears.count_file_inputs(gear) != 1:
-        raise Exception("Legacy gear enqueue attempt of " + gear_id + " failed: must have exactly 1 input in manifest")
+    if gears.count_file_inputs(gear) - len(fixed_input_keys) != 1:
+        # This shouldn't happen if the handler is correctly validating the POST and PUT methods for rules
+        log.error("Legacy gear enqueue attempt of " + gear_id + " failed: must have exactly 1 input non-fixed input from the manifest, it has {} non-fixed inputs".format(gears.count_file_inputs(gear) - len(fixed_input_keys)))
```
ehlertjd: `exactly 1 input non-fixed input` -> `exactly 1 non-fixed input`

api/jobs/rules.py
```diff
@@ -355,13 +350,47 @@ def validate_regexes(rule):
...
+    fixed_inputs = fixed_inputs if fixed_inputs else []
+    if gears.count_file_inputs(gears.filter_optional_inputs(geardoc)) - len(fixed_inputs) != 1:
+        raise InputValidationException("Rule must have exactly 1 non-fixed gear value")
+    for fixed_input in fixed_inputs:
+        if not geardoc['gear']['inputs'].get(fixed_input['input']):
+            raise InputValidationException("Unrecognized gear input cannot be fixed for the rule")
+        # Check to see that each fixed input actually exists
+        create_filereference_from_dictionary(fixed_input).get_file()
+
+class Rules:
```
ehlertjd: Love it! From a structure standpoint, I would prefer that this to end up in a `mappers` submodule and have docstrings.

api/jobs/jobs.py
hkethi002: The check happens in enqueue.py

nagem:
>* All optional inputs are still ignored.

Are you able to set a fixed input for optional inputs?

hkethi002:
Nope, as of now optional inputs are always ignored by rules, and I wasn't sure if we wanted to keep it that way, use the policy stuff we implemented for batch runs, or some other option.

api/handlers/listhandler.py
hkethi002: Since in 99% of case, the project should be the project the file is in, I think the project id should be fine, I am giving the rule name.

hkethi002:
I am going to create a follow up ticket/PR to refactor the rest of rules code to use the mapper and model.

api/jobs/handlers.py
```diff
@@ -393,6 +398,10 @@ def post(self, cid):
         if requires_read_write_key(get_gear(payload['gear_id'])):
             raise InputValidationException("Cannot create rule with a gear that requires a read-write api-key.")
 
+        # Site rules can't have fixed_inputs
+        if payload.get('fixed_inputs') and cid == 'site':
+            raise InputValidationException("Cannot create a site rule with a fixed input/")
```
ehlertjd: extra trailing slash

jody-zeitler:
Frontend approved!

PR #1517, opened 2019-01-29
[FLYW-348] Age should remain on session level, with DB Fix

ehlertjd:
Revisiting #1512 - With a new process for applying database fixes that are not schema changes.

The new upgrade flow is:
- If there are any outstanding fixes for the current DB version, apply them
- Then upgrade to the next schema version and apply any outstanding fixes.
- Repeat until fully upgraded & all fixes have been applied.

Applied fixes are stored in the `version` singleton as a map of fix ID to date applied. 

This makes this PR mergeable to both hotfix branches and master with minimal conflicts.

### Review Checklist


tests/integration_tests/python/test_upgrades.py
```diff
@@ -1004,24 +997,6 @@ def test_62(api_db, data_builder, database, default_payload, as_admin, file_form
...
-
-        # Job with empty dictionary input
-        job_data = api_db.jobs.find_one({'_id': job_empty_dict_input})
-        assert job_data['parents']['group'] == group
-        assert job_data['parents']['project'] == bson.ObjectId(project)
-        assert job_data['parents']['subject'] == bson.ObjectId(subject)
-        assert job_data['parents']['session'] == bson.ObjectId(session)
-        for cid in dest_containers:
-            assert cid in job_data['related_container_ids']
-
```
hkethi002: Why are we removing these?

ehlertjd: Good question, I'll amend

mjburling:
@ehlertjd ignore me if this has already been shot down, but I'm a fan of [PEP 257](https://www.python.org/dev/peps/pep-0257/). The modules and functions look naked without docstrings.

bin/fixes.py
```diff
@@ -0,0 +1,46 @@
+
+from api import config
+
+from process_cursor import process_cursor
+
+# NOTE: When adding a fix, create a new list for the current
+# DB version (if it doesn't exist) and add your entry to the back.
+# That fix will be applied before upgrading from that version to the next.
+AVAILABLE_FIXES = {
+    62: [ 'fix_subject_age_62' ]
```
mjburling: I think this pattern is really simple and sorely needed.
🌟 🍰

mjburling: But, maybe the NOTE should live in a module-level docstring?

bin/process_cursor.py
```diff
@@ -0,0 +1,74 @@
...
+
+    Useful for upgrades that need to touch each document in a database, and don't need an iteration order.
+
+    Your closure MUST return True on success. Anything else is logged and treated as a failure.
+    A closure that throws an exception will fail the upgrade immediately.
+    """
+
+    begin = get_monotonic_time()
+
+    # cores = multiprocessing.cpu_count()
```
mjburling: Are these lines of commented code meaningful?

ehlertjd: Included for legacy reasons.

ehlertjd:
@mjburling Docstrings added.

bin/database.py
```diff
@@ -25,19 +25,44 @@
...
+        list(str): The list of fix_ids to be applied
+    """
+    available_fixes = fixes.AVAILABLE_FIXES.get(db_version, [])
+
+    result = []
+    for fix in available_fixes:
+        if fix not in applied_fixes:
+            result.append(fix)
+
+    return result
```
kofalt: This function should probably error out if a fix is not present in the DB, for a lower version that the DB is currently at. Or if not here, definitely somewhere - I don't see right now where that would happen if one is skipped.

bin/database.py
```diff
@@ -2389,14 +2345,35 @@ def upgrade_to_62():
...
+
+    Args:
+        db_version (int): The current database version
+        applied_fixes (dict): The map of fixes that have already been applied
+        update_doc (dict): The document to update with applied fixes.
+    """
+    available_fixes = get_available_fixes(db_version, applied_fixes)
+
+    for fix_id in available_fixes:
+        fix_fn = getattr(fixes, fix_id)
```
kofalt: I would hope for a cleaner way to get this out than `getattr` - maybe a getter in the package that in turn calls getattr to place it next to the funcs it references. Or, `AVAILABLE_FIXES` could hold function pointers instead of string literals, and just use `__name__` where appropriate.

ehlertjd:
@mjburling Now targeting `master` branch

kofalt:
I like what I see. Still not reviewing `fix_subject_age_62`, defer to others

bin/fixes.py
```diff
@@ -0,0 +1,130 @@
...
+    result = globals().get(fix_id, None)
+    if not result:
+        raise ValueError('Unknown fix method: {}'.format(fix_id))
+    return result
+
+def set_session_age_from_job(session, subject_age):
+    classifier_job = list(config.db.jobs.find({'parents.session': session['_id'], 'gear_info.name': {'$regex': 'classifier$'}, 'produced_metadata.session.subject.age': {'$exists': True}}).limit(1).sort('created', direction=-1))
+    if not classifier_job and session.get('age') is None:
+        config.log.warning('Could not find classifier job for session {} that produced a Patient Age, using subject age even though there are multiple sessions for the subject'.format(session['_id']))
+        config.db.sessions.update({'_id': session['_id']}, {'$set': {'age': subject_age}})
```
hkethi002: In the case that we don't find a classifier job for the session, we update the session age with the subject age, that may not be appropriate since, if the classifier never ran on it, it would not have set the age. 
If that is the case, I can reduce the log level to an info level and not update the session with the subject age field.

@ehlertjd @mjburling @nagem

ehlertjd: In at least some cases, we found that the warnings were accurate due to us not preserving actual `produced_metadata` on the job object. So @hkethi002 is going to modify the fix logic such that:
1. If the most recent classifier job doesn't have a `subject.age` in the `produced_metadata`, we'll read the `PatientAge` header and perform the calculation.
2. If there are no classifier jobs, we'll ignore that session.

Finally, we should update the placer logic such that the `raw_subject` is set on the session, but the job preserves the **original** metadata.

nagem: >due to us not preserving actual produced_metadata on the job object

What were we doing instead?

ehlertjd: We are currently preserving the modified metadata (i.e. after removing `subject` and replacing it with `session.info.subject_raw`)

bin/database.py
```diff
@@ -2426,32 +2367,42 @@ def upgrade_schema(force_from = None):
...
+        config.db.singletons.update_one({'_id': 'version'}, {'$set': update_doc})
         sys.exit(0)
 
 if __name__ == '__main__':
     try:
         parser = argparse.ArgumentParser()
         parser.add_argument("function", help="function to be called from database.py")
         parser.add_argument("-f", "--force_from", help="force database to upgrade from previous version", type=int)
+        parser.add_argument("-F", "--fix-id", help="ID of fix to run if applying a fix")
         args = parser.parse_args()
```
nagem: It's a bit confusing to have a `-f` and `-F` that do mostly unrelated things but not the end of the world.

bin/fixes.py
```diff
@@ -94,14 +94,45 @@ def get_fix_function(fix_id):
...
+        'D': 1,
+    }
+    scale = age[-1:]
+    value = age[:-1]
+    if scale not in conversion.keys():
+        # Assume years
+        scale = 'Y'
+        value = age
+
+    age_in_seconds = datetime.timedelta(int(value) * conversion.get(scale)).total_seconds()
```
ehlertjd: Catch `ValueError` and warn if `value` is a non-int?

bin/fixes.py
```diff
@@ -94,14 +94,45 @@ def get_fix_function(fix_id):
...
+
+    # Make sure that the age is reasonable
+    if not age_in_seconds or age_in_seconds <= 0:
+        age_in_seconds = None
+
+    return age_in_seconds
+
 def set_session_age_from_job(session, subject_age):
-    classifier_job = list(config.db.jobs.find({'parents.session': session['_id'], 'gear_info.name': {'$regex': 'classifier$'}, 'produced_metadata.session.subject.age': {'$exists': True}}).limit(1).sort('created', direction=-1))
+    classifier_job = list(config.db.jobs.find({'parents.session': session['_id'], 'gear_info.name': {'$regex': 'classifier$'}, 'produced_metadata.acquisition.metadata.PatientAge': {'$exists': True}}).limit(1).sort('created', direction=-1))
```
ehlertjd: I think I would rather we get the latest classifier job, and make a decision between taking `subject.age` if present and falling back to this approach if not present. a. la. [this comment](https://github.com/flywheel-io/core/pull/1517#discussion_r252332094)

ehlertjd: As written, assumes that the only classifiers are dicom classifiers - maybe that's true?

bin/fixes.py
```diff
@@ -94,14 +94,45 @@ def get_fix_function(fix_id):
...
     if not classifier_job and session.get('age') is None:
-        config.log.warning('Could not find classifier job for session {} that produced a Patient Age, using subject age even though there are multiple sessions for the subject'.format(session['_id']))
-        config.db.sessions.update({'_id': session['_id']}, {'$set': {'age': subject_age}})
+        config.log.warning('Could not find classifier job for session {} that produced a Patient Age'.format(session['_id']))
```
ehlertjd: This would still warn in the case where there were no DICOM Acquisitions on a session - I think we can adopt the previously discussed tact of just printing an `info` message and leaving the session untouched.

bin/database.py
nagem: Thinking about it a little longer .. if you got it wrong it would be a fast failure. Only worry would be if it was a cryptic/harmful/silent failure. Let's just keep it as is.

bin/fixes.py
```diff
@@ -125,12 +130,18 @@ def parse_patient_age(age):
...
         classifier_job = classifier_job[0]
-        session_age = parse_patient_age(classifier_job['produced_metadata']['acquisition']['metadata']['PatientAge'])
+        if classifier_job.get('produced_metadata', {}).get('session', {}).get('subject', {}).get('age'):
+            session_age = classifier_job['produced_metadata']['session']['subject']['age']
+        elif classifier_job.get('produced_metadata', {}).get('acquisition', {}).get('metadata', {}).get('PatientAge'):
+            session_age = parse_patient_age(classifier_job['produced_metadata']['acquisition']['metadata']['PatientAge'], job_id=classifier_job['_id'])
+        else:
+            log.info('Job {} did not produce subject age or PatientAge')
+            return True
         config.db.sessions.update({'_id': session['_id']}, {'$set': {'age': session_age}})
```
ehlertjd: Do we need to check whether or not `subject_age` is `None` before running the update (And log an `info` statement)?

hkethi002: I think we need to check, however, since the only case I can see it being `None` is when we catch the ValueError in `parse_patient_age`, which we log anyways, logging it may clutter up the logs.

ehlertjd:
My last concerns regarding the fix itself have been addressed. Worth another spin @mjburling?

bin/fixes.py
hkethi002: There is an edge where the latest classifier job has neither `produced_metadata.session.subject.age` nor `produced_metadata.acquisition.metadata.PatientAge` when it is run again because metadata is only produced on fields that are null/empty. So I think getting the latest job with at least the PatientAge is the correct query to make.

bin/fixes.py
```diff
@@ -128,18 +128,23 @@ def parse_patient_age(age, job_id=None):
...
-    elif classifier_job:
-        classifier_job = classifier_job[0]
-        if classifier_job.get('produced_metadata', {}).get('session', {}).get('subject', {}).get('age'):
-            session_age = classifier_job['produced_metadata']['session']['subject']['age']
-        else:
-            session_age = parse_patient_age(classifier_job['produced_metadata']['acquisition']['metadata']['PatientAge'], job_id=classifier_job['_id'])
+def set_session_age_from_file_info(session, subject_age):
+
+    # Get all the files in each acquisition
+    list_of_file_lists = list(map((lambda a: a.get('files', [])), list(config.db.acquisitions.find({'session': session['_id']}))))
```
nagem: Could you just make the mongo query use an `elemMatch` query where `'files.type': 'dicom' ` and `$exists: {'files.info.PatientAge': True}`?

ehlertjd: I think we still can't assume `dicom`

hkethi002: We would still need the filter at the end though since that won't get rid of extra files in the acquisition though, right?

mjburling:
I think we're generally there. I'd like to have an explicit approval from @ryansanford for a handful of sites he'll be running against this morning.

@ehlertjd @hkethi002 The previous upgrades (maybe for 62?) were displaying those CRIT messages, but I haven't seen any since this specific branch. Did some assumptions change that negate those old CRIT messages? I don't think I have any logs locally that exhibit that behavior, but may @ryansanford could share the ones he had been seeing if they come up again?

hkethi002:
The WARN level messages were reduced to INFO level; we decided that for sessions that had no files with PatientAge in the header, or any Acquisition metadata, there wasn't anything we could do short of downloading the file and checking the header.

PR #1520, opened 2019-01-30
Add /jobs/ask

kofalt:
Adds a new route `/jobs/ask`  - ask the job queue a question.
Does away with the flak from `/jobs/next` and `/jobs/stats` to unify job filtering under a common struct. 

Along the way, several function signatures got a lot more readable. Moved existing calls around a bit and made them call the new hotness. Also removed `/jobs/pending` on the basis that it's never been used.

Tests / docs / schema TBD after approach is reviewed.
Multi-job support not in scope for this PR as a later addition. Func sigs support it.

Adds a deep_equals helper, that lives on `QueueTestCases` for now. Useful output:

```
# stdout capture
Result:   {'tags': {'$in': ['example-tag']}}
Expected: {'tags': {'$nin': ['example-tag']}}

# stderr capture
{'dictionary_item_removed': set(["root['tags']['$in']"]), 'dictionary_item_added': set(["root['tags']['$nin']"])}
```

### DIFF
--- a/api/api.py
+++ b/api/api.py
@@ -148,9 +148,9 @@ endpoints = [
 
         route( '/jobs',                             JobsHandler),
         prefix('/jobs', [
+            route('/ask',                           JobsHandler, h='ask',                  m=['POST']),
             route('/next',                          JobsHandler, h='next',                 m=['GET']),
             route('/stats',                         JobsHandler, h='stats',                m=['GET']),
-            route('/pending',                       JobsHandler, h='pending',              m=['GET']),
             route('/reap',                          JobsHandler, h='reap_stale',           m=['POST']),
             route('/add',                           JobsHandler, h='add',                  m=['POST']),
             route('/<:[^/]+>',                      JobHandler),
--- a/api/jobs/queue.py
+++ b/api/jobs/queue.py
@@ -16,7 +16,7 @@ from ..dao.containerutil import (
     create_containerreference_from_filereference, FileReference
 )
 from .job_util import resolve_context_inputs
-from ..web.errors import InputValidationException
+from ..web.errors import InputValidationException, APIValidationException
 
 
 log = config.log
@@ -397,7 +397,7 @@ class Queue(object):
         result = {}
 
         if jobs <= 0 and not stats:
-            raise InputValidationException('Not asking for work or stats')
+            raise APIValidationException('Not asking for work or stats')
 
         if jobs > 0:
             result['jobs'] = Queue.start_jobs(jobs, query['whitelist'], query['blacklist'], query['capabilities'], peek)
@@ -407,7 +407,7 @@ class Queue(object):
         return result
 
     @staticmethod
-    def lists_to_query(whitelist, blacklist):
+    def lists_to_query(whitelist, blacklist, capabilities):
         """
         Translate a whitelist and blacklist to job database query.
         """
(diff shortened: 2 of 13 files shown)


api/jobs/queue.py
```diff
@@ -361,29 +361,142 @@ def enqueue_job(job_map, origin, perm_check_uid=None):
...
+        if match['tag']:
+            query['tags'] = match['tag']
+
+        log.info('Job query is: %s', json.dumps(query))
+        return query
+
+    @staticmethod
+    def start_jobs(max_jobs, whitelist, blacklist, capabilities, peek):
+        """
+        Atomically change up to N jobs from pending to running.
```
ehlertjd: For all `N=[1,2)`?

api/jobs/queue.py
```diff
@@ -361,29 +361,142 @@ def enqueue_job(job_map, origin, perm_check_uid=None):
...
+
+        Will return empty array if there are no jobs to offer. Searches for jobs in FIFO order.
+        """
+
+        if max_jobs > 1:
+            raise InputValidationException('Starting multiple jobs not supported')
+        if max_jobs < 1:
+            raise InputValidationException('Must start at least one job')
+        if peek and max_jobs > 1:
+            raise InputValidationException('Cannot peek more than one job')
```
ehlertjd: Assuming that these checks will be updated when multiple job support is added, as of today the last check is unreachable.

requirements.txt
```diff
@@ -19,4 +19,5 @@ unicodecsv==0.9.0
 urllib3==1.23
 webapp2==2.5.2
 WebOb==1.8.2
+deepdiff>=3.3.0
```
ehlertjd: 📎: It looks like this should go in `tests/requirements.txt`

api/api.py
```diff
@@ -148,9 +148,9 @@ def prefix(path, routes):
 
         route( '/jobs',                             JobsHandler),
         prefix('/jobs', [
+            route('/ask',                           JobsHandler, h='ask',                  m=['POST']),
             route('/next',                          JobsHandler, h='next',                 m=['GET']),
             route('/stats',                         JobsHandler, h='stats',                m=['GET']),
-            route('/pending',                       JobsHandler, h='pending',              m=['GET']),
```
ehlertjd: Which clients use this?

api/jobs/queue.py
kofalt: > Multi-job support not in scope for this PR as a later addition. Func sigs support it.

Wanted to get the structure in and not bother with testing multi-job until later, since group filtering is \~urgent~

api/api.py
kofalt: None! I added it once upon a time then never used it

requirements.txt
kofalt: Solid, moving

kofalt:
> I do wonder if there's a point where we codify these rulesets in the database rather than ad-hoc on a per client basis.

There is: the compute dipsatch design does this, and enables self-service. Ideally this same struct will be used

kofalt:
Should be ready for merge.

api/jobs/queue.py
```diff
@@ -361,29 +361,141 @@ def enqueue_job(job_map, origin, perm_check_uid=None):
...
+        jobs  = query['return'].get('jobs',   0)
+        stats = query['return'].get('states', False)
 
-        inclusive_tags = filter(lambda x: not x.startswith('!'), tags)
-        exclusive_tags =  map(lambda x: x[1:], filter(lambda x: x.startswith('!'), tags)) # strip the '!' prefix
+        result = {}
 
-        query = { 'state': 'pending' }
+        if jobs <= 0 and not stats:
+            raise InputValidationException('Not asking for work or stats')
```
ehlertjd: `APIValidationException` would be mildly more appropriate

api/jobs/queue.py
```diff
@@ -361,29 +361,141 @@ def enqueue_job(job_map, origin, perm_check_uid=None):
...
+
+        # Translate to mongo keys
+        if match['group']:
+            query['parents.group'] = match['group']
+        if match['gear-name']:
+            query['gear_info.name'] = match['gear-name']
+        if match['tag']:
+            query['tags'] = match['tag']
+
+        log.info('Job query is: %s', json.dumps(query))
```
ehlertjd: Can this be a `debug` statement?

api/jobs/queue.py
kofalt: Addressed

api/jobs/queue.py
kofalt: Addressed

PR #1521, opened 2019-01-31
Save actual produced_metadata

ehlertjd:
We manipulate engine metadata before saving it as produced_metadata.
This results in a potential loss in fidelity of produced_metadata in the database.
Now we make a copy to save before we start manipulating the metadata.

See #1517 

### Review Checklist


tests/integration_tests/python/test_uploads.py
```diff
@@ -1069,6 +1069,13 @@ def test_acquisition_engine_upload(data_builder, file_form, as_root):
     )
     assert r.ok
 
+    expected_metadata = copy.deepcopy(metadata)
+    expected_metadata['acquisition']['files'][2]['name'] = 'folderB/two.csv'
```
hkethi002: Why is it `'folderB/two.csv'`? Shouldn't it be `'../folderB/two.csv'` since that is the raw metadata that is being uploaded?

ehlertjd: That gets trickier as we're manipulating filenames way ahead of this logic, and would require a more significant change. To me it didn't seem quite as worth it, because then you'd end up with mismatched paths between `produced_metadata` and what actually gets saved on the hierarchy.

PR #1525, opened 2019-02-04
Quick stats patch for exclusive tags

kofalt:
It looks like `/job/stats` does not consider exclusive tags - those prefixed with `!`.
This requires multiple round-trips to get exclusive stats, when it should be just one.

As a quick patch, I copied the `start_job` behavior to `get_statistics`.
Makes it ever-clearer that tag filtering should be removed; this effort will be obsoleted by #1520.

### DIFF
--- a/api/jobs/queue.py
+++ b/api/jobs/queue.py
@@ -504,8 +504,6 @@ class Queue(object):
         inclusive_tags = filter(lambda x: not x.startswith('!'), tags)
         exclusive_tags =  map(lambda x: x[1:], filter(lambda x: x.startswith('!'), tags)) # strip the '!' prefix
 
-        query = { 'state': 'pending' }
-
         if len(inclusive_tags) > 0 or len(exclusive_tags) > 0:
             query['tags'] = {}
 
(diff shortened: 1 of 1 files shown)


api/jobs/queue.py
```diff
@@ -496,13 +496,30 @@ def get_statistics(tags=None, last=None, unique=False, all_flag=False):
...
+
+        if len(inclusive_tags) > 0:
+            query['tags']['$in']  = inclusive_tags
+        if len(exclusive_tags) > 0:
+            query['tags']['$nin'] = exclusive_tags
+
 
         if tags is not None and len(tags) > 0:
-            match = { 'tags': {'$in': tags } } # match only jobs with given tags
+            query = { 'tags': {'$in': tags } } # match only jobs with given tags
```
ehlertjd: Don't override query after setting it above.

kofalt: You're no fun at all!

ehlertjd:
Functionally looks fine - we could share the query-building code as a function between stats/next and appease `codecov`, but in light of #1520 I don't care.

kofalt:
Merging since this doesn't seem to actively break anything, and is shown to be better than the state of play on at least one test site. Should be replaced by #1520 shortly thereafter.

PR #1526, opened 2019-02-05
[FLYW-368] Add build-your-own tree endpoint

ehlertjd:
This new endpoint allows API consumers to request a portion of the
hierarchy that they care about, returning only the relevant fields.

Permits top-level pagination, and filtering, sorting and limiting of
joined collections.

Operates on the full hierarchy, including analyses and jobs!

See examples in [integration tests](https://github.com/flywheel-io/core/blob/tree-endpoint/tests/integration_tests/python/test_tree.py#L191-L198).

### Review Checklist


nagem:
This is really great, Justin. Can't wait to replace all calls in the clients with this! 😉

api/api.py
```diff
@@ -366,6 +367,10 @@ def prefix(path, routes):
         route('/views/data', DataViewHandler, h='execute_adhoc', m=['POST']),
         route('/views/columns', DataViewHandler, h='get_columns', m=['GET']),
 
+        # Tree-based retrieval
+        route('/tree',       TreeHandler,                m=['PUT']),
```
nagem: Any reason this is a `PUT` rather than a `POST`? We've used `POST` historically for retrievals that need to support a request body but I was curious your reasoning.

ehlertjd: No particular reason - I'll likely switch this out in the final version.

ehlertjd:
@nagem I haven't added support for collections, but seems like it would be possible.
*However*, we don't support adding subjects to collections AFAIK.

nagem:
We don't support adding subjects to collections, I'm just imagining a Subject View while on a collection. Currently the only thing you can actually add to collections is acquisitions. I would likely just add a `get_subjects_for_collection` endpoint to the collection handler and be done with it if that is needed.

nagem:
After a re-review this looks good to go 👍 . 

If you'd like to keep the debug log statements that's fine, but this is just a reminder that there are ~5 of them scattered around.

PR #1529, opened 2019-02-07
[FLYW-401] Abstract Container search for group with 12 character id 404s

hkethi002:
### Review Checklist


ehlertjd:
@hkethi002 What I was getting at was I think you could just add `len(cid) == 24` to the [original check](https://github.com/flywheel-io/core/pull/1529/files#diff-2ac58d64895c18245a2a5c5c8f8b84a5L22)

hkethi002:
But what if there's a group id that's 24 characters long?

api/dao/containerutil.py
```diff
@@ -376,9 +376,22 @@ def container_search(query, projection=None, collections=PLURAL_CONT_TYPES, earl
...
+            bson_query['_id'] = bson.ObjectId(bson_query['_id'])
+        else:
+            bson_query = None
+
+
     for coll_name in collections:
         coll = config.db.get_collection(coll_name)
-        coll_results = list(coll.find(query, projection, **kwargs))
+        coll_results = []
+        if coll_name in ['groups', 'users'] and query.get('_id'):
```
ehlertjd: Any way we can NOT hardcode the set of collections? (e.g. use [use_object_id](https://github.com/flywheel-io/core/blob/master/api/dao/basecontainerstorage.py#L47) on ContainerStorage)

PR #1530, opened 2019-02-07
[FLYW-402] Site settings redux

ehlertjd:
This is ready for review.

Adds [site level settings](https://flywheel-io.github.io/core/branches/site-settings-redux/swagger/index.html#/site/get_site_settings) which allow the registration of center-pays gears. 

### Review Checklist

- Extra care is taken and l)og messages are created for [Critical Data](https://github.com/flywheel-io/core/wiki/Critical-Data-Flows)

### DIFF
--- a/api/site/handlers/site_settings_handler.py
+++ b/api/site/handlers/site_settings_handler.py
@@ -1,0 +1,27 @@
+from ...web import base
+from ...auth import require_admin, require_login
+from ... import validators
+
+from ..mappers import SiteSettingsMapper
+
+class SiteSettingsHandler(base.RequestHandler):
+    """Handler for admin editable site-wide configuration"""
+
+    @require_login
+    def get(self):
+        """Return site settings"""
+        mapper = SiteSettingsMapper()
+        settings = mapper.find()
+        return settings.to_dict() if settings else {}
+
+    @require_admin
+    @validators.verify_payload_exists
+    def put(self):
+        """Patch site setting values"""
+        # Validate
+        payload = self.request.json
+        validators.validate_data(payload, 'site-settings.json', 'input', 'PUT')
+
+        # Update/upsert
+        mapper = SiteSettingsMapper()
+        mapper.patch(payload)
--- a/api/site/handlers/site_settings_handler.py
+++ b/api/site/handlers/site_settings_handler.py
@@ -1,8 +1,9 @@
+"""Provides handler for site settings endpoint"""
 from ...web import base
 from ...auth import require_admin, require_login
 from ... import validators
 
-from ..mappers import SiteSettingsMapper
+from ..site_settings import get_site_settings, update_site_settings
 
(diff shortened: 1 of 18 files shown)


api/site/handlers/site_settings_handler.py
```diff
@@ -0,0 +1,26 @@
...
+    @require_admin
+    @validators.verify_payload_exists
+    def put(self):
+        """Patch site setting values"""
+        # Validate Input
+        payload = self.request.json
+        validators.validate_data(payload, 'site-settings.json', 'input', 'PUT')
+
+        # Update/upsert
+        update_site_settings(payload)
```
roycollins: Could we return a 422 on InvalidInput exceptions or does it always throw just the top level 400 on exceptions?

ehlertjd: I think 422 is appropriate for semantic validation - will update, thanks!

swagger/schemas/definitions/config.json
```diff
@@ -93,6 +93,31 @@
...
+        },
+
+        "site-settings": {
+            "type": "object",
+            "properties": {
+                "created": {"type": "string"},
+                "modified": {"type": "string"},
+                "center_gears": {
+                    "type": "array",
+                    "items": {"type": "string"},
```
hkethi002: We can use `{"$ref": "gear.json#/definitions/gear-name"}` to validate that the string is a valid gear name, but maybe it isn't necessary with the 422 that's raised anyways.

swagger/schemas/definitions/config.json
```diff
@@ -93,6 +93,31 @@
...
+                }
+            },
+            "required": ["created", "modified", "center_gears"]
+        },
+        "site-settings-input": {
+            "type": "object",
+            "properties": {
+                "center_gears": {
+                    "type": "array",
+                    "items": {"type": "string"},
```
hkethi002: Same as above

api/site/site_settings.py
```diff
@@ -0,0 +1,49 @@
+from .models import SiteSettings
+from .mappers import SiteSettingsMapper
+from ..jobs import gears
+from ..web import errors
```
kofalt: What would you think about grouping these methods as static under a class?

ehlertjd: I debated that - from an extensibility and testability standpoint, I think it does make more sense than functions on a module. The main goal with this submodule was to start defining a Domain API for internal consumption across package boundaries.

ehlertjd: Going back and forth on this, and I think I'm going to punt for now. On one hand @ambrussimon had a good point about having to instantiate mappers via constructor. The reason I made the decision on mappers is that there may be cases where you want your mapper instance to be isolated. (e.g. you're talking to a different data source, or using an identity map)

For the interface to the domain layer a top level facade makes sense to me, and it seems that a module with "static" functions is more idiomatic than a class with static functions.

PR #1531, opened 2019-02-08
Index job created

hkethi002:
- Jobs view requires admins to be able to get logs without needing permissions or orange mode
- Sorting a bunch of jobs by created timestamp needs it to be indexed
### Review Checklist


api/jobs/handlers.py
```diff
@@ -803,7 +803,7 @@ def _log_read_check(self, _id):
             self.abort(404, 'Job not found')
 
         # Permission check
-        if not self.superuser_request:
+        if not self.user_is_admin:
```
ehlertjd: How about some #1296?

api/jobs/handlers.py
```diff
@@ -803,7 +803,7 @@ def _log_read_check(self, _id):
...
             if job.inputs is not None:
```
ehlertjd: We may need to update at least one unit test to get coverage of this section back - we're probably just creating all jobs `as_admin`.

hkethi002: Yeah, I was going to test it with a user with permissions and without

PR #1532, opened 2019-02-08
[FLYW-405] File abstraction layer 

roycollins:
Creates a file abstraction layer that interacts between the flywheel code and any current storage provider we support; local files and GCP currently testing

Ready for review and full testing

### Review Checklist


api/files.py
```diff
@@ -290,13 +312,15 @@ def get_fs_by_file_path(file_path):
     still supports the legacy storage, attempt to serve from there.
     """
 
-    if config.fs.isfile(file_path):
-        return config.fs
-    elif config.support_legacy_fs and config.local_fs.isfile(file_path):
+    if config.py_fs._fs.isfile(file_path):
```
ehlertjd: I think we could invert the order of checks in this function, and skip the expensive GCS `isfile` check. Thoughts @nagem?

api/placer.py
```diff
@@ -583,20 +590,32 @@ def finalize(self):
             'id': uid
         }
 
+        # Finaly move the file from the temp fs to the persistent FS. 
+        # We could make this faster using a move if we know its a local to local fs move.
+        with self.file_processor.temp_fs.open(None, tempZipPath, 'rb', None) as (f1
+                ), self.file_processor.persistent_fs.open(token, util.path_from_uuid(token), 'wb', None) as f2:
+            for line in f1:
```
ehlertjd: Prefer chunked copying, in ~1MB chunks. I'm not sure what `readline` will do for a binary file.

api/storage/__init__.py
```diff
@@ -0,0 +1,2 @@
+def create_flywheel_fs(url):
```
ehlertjd: Shouldn't we use this in `config.py`? :)

api/storage/py_fs_file.py
```diff
@@ -0,0 +1,59 @@
+from .storage import Storage
+from fs import open_fs
+import fs
+
+#This is OSFS only.   Which is legacy only
+
+class PyFsFile(Storage):
```
ehlertjd: Possibly rename this class to `PyFsStorage` after it's parent class?

api/storage/storage.py
```diff
@@ -0,0 +1,36 @@
+from abc import ABCMeta, abstractmethod
+
+class Storage(object):
+    """Abstract class for filessytem objects"""
```
ehlertjd: Please document the full class, preferably using [Google Style Docstrings](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html)

api/upload.py
```diff
@@ -81,17 +81,28 @@ def process_upload(request, strategy, access_logger, container_type=None, id_=No
...
         name_fn = util.sanitize_path
     else:
         name_fn = os.path.basename
 
+
     # The vast majority of this function's wall-clock time is spent here.
-    # Tempdir is deleted off disk once out of scope, so let's hold onto this reference.
-    file_processor = files.FileProcessor(config.fs, local_tmp_fs=(strategy == Strategy.token), tempdir_name=tempdir)
+    local_tmp_fs = None
+    needsTempFile = ['token', 'packfile']
```
ehlertjd: We could store this as an attribute on the `Strategy` class

api/upload.py
```diff
@@ -130,15 +142,22 @@ def process_upload(request, strategy, access_logger, container_type=None, id_=No
...
+
+        #Some placers need this value. Consistent object would be nice
+        field.path = filePath
+
+        # we can get the size from the file on disk but we need to know which fs its stored on.
+        # TODO: we should make it part of the Storage abstraction layer
+        if file_processor._temp_fs:
+            field.size = int(file_processor._temp_fs._fs.getsize(filePath))
+        else:
+            field.size = int(file_processor._persistent_fs._fs.getsize(filePath))
```
ehlertjd: Do we need a `get_file_info` interface on the Storage interface that returns a few attributes? (e.g. exists, file size & hash)
We definitely shouldn't be reaching into the `_fs` attribute, since that won't always be there.

api/config.py
```diff
@@ -7,6 +7,7 @@
 import elasticsearch
 
 from fs import open_fs
```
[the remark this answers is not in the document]
roycollins: Yes, should be able to. I was not sure if removing the definition from this scope would effect other files that include config and then rely on the import to still be there.  Does python scope per file or does it inherit down on children includes?  I didnt want to break other code that might use this.

api/config.py
```diff
@@ -309,8 +310,11 @@ def get_release_version():
 
 # Storage configuration
 fs = open_fs(__config['persistent']['fs_url'])
```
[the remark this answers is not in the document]
roycollins: Correct.  I just kept it in place so I could swap between testing vals while I was debugging. Ill remove it once we finalize on what we want to call the new variable

api/config.py
```diff
@@ -309,8 +310,11 @@ def get_release_version():
...
-local_fs = open_fs('osfs://' + __config['persistent']['data_path'])
+#local_fs = open_fs('osfs://' + __config['persistent']['data_path'])
+local_fs = PyFsFile('osfs://' + __config['persistent']['data_path'])
```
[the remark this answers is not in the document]
roycollins: If we agree that local_fs will always be on the local disk that would be fine. We should add a type check to confirm that it is always os type too.  I assume os works the same using local disks, nas, san, NVMe, and any type of local block storage. I just dont want to make any assumptions from my limited python knowledge

api/config.py
```diff
@@ -309,8 +310,11 @@ def get_release_version():
...
 support_legacy_fs = __config['persistent']['support_legacy_fs']
+py_fs = PyFsFile(__config['persistent']['fs_url'])
+fs = py_fs
```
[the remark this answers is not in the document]
roycollins: I think storage is a good name too.  I will get these swapped over.

api/config.py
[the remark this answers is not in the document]
ehlertjd: Yeah we should be safe in assuming that any such storage would be mounted and presented as a regular filesystem accessible to python - historically that's how it has always worked for local storage AFAIK.

api/create_file.py
```diff
@@ -1,93 +0,0 @@
...
-        self.path = path
-        self.fileobj = fileobj
-        
-        if metadata is not None:
-            self.metadata = metadata.copy()
-        else:
-            self.metadata = {}
-        self.metadata['name'] = filename
-
-class FileCreator(object):
```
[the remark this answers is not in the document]
roycollins: Most of the code was duplicated in the process_file within files.py as well.  I moved all file processing to the same service class.  There are about 15 lines that are unique to the TargetedMulitPlacer only.  Those lines were inlined in the 2 places they were used. Most of those 15 lines could be refactored into a common function in files.py but it feels like the placer should know internal how its going to process files. I just wanted to talk it over before forcing a decision myself.

api/data_views/file_opener.py
```diff
@@ -54,7 +54,7 @@ def __enter__(self):
         try:
             # Open the file using file_system
             file_path, file_system = files.get_valid_file(self.file_entry)
-            self._system_fd = file_system.open(file_path, open_mode)
+            self._system_fd = file_system.open(None, file_path, open_mode, None)
```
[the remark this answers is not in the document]
roycollins: Future ones will require the uuid, yes.  Ill throw them in there if we have it.

api/files.py
```diff
@@ -16,47 +15,34 @@
...
 
-        Returns:
-            str, file: The path and opened file
+        self._persistent_fs = persistent_fs
+        self._temp_fs = None
+        self._tempdir_name = tempdir_name
+
+        if local_tmp_fs:
+            self._temp_fs = local_tmp_fs
+            if not self._temp_fs._fs.exists(tempdir_name):
```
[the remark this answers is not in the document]
roycollins: I agree completely.

api/files.py
```diff
@@ -274,9 +292,13 @@ def get_file_path(file_info):
 
 
 def get_signed_url(file_path, file_system, **kwargs):
+    """ 
+    @deprecated
```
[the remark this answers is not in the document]
roycollins: Yes,  It is used in a few places and I didnt want to existing code.  Specifically the jobs.py since I am not sure how to even run a job to test that its not broken.  It can be quick refactor

api/files.py
roycollins: I think the best refactor would be to store the fs type on the file.  Then it would be known without any work.  I also think we should move files to a first class object but that will be a larger refactor.

api/placer.py
roycollins: Im not sure if this is calling readline under the hood.  I think its just piping the byte streams together but I am not sure.  I can do chunking easily enough if this is calling readline internally.

api/storage/__init__.py
roycollins: heh, Yes we should.  I completely forgot about this special init function.

api/storage/py_fs_file.py
roycollins: Will do

api/storage/storage.py
roycollins: Will do

api/upload.py
roycollins: I agree but the Strategy class is just an enum I believe.  I am not sure if we  Enum allows more than key/value.  We could convert it to a full object if needed, I would say preferable.

api/upload.py
roycollins: I think we do.  I threw the note in there to check with you to confirm.  I would say yes just so we dont have to reach inside the interface boundry.

api/jobs/handlers.py
```diff
@@ -325,11 +325,14 @@ def download(self, **kwargs): # pragma: no cover
             '_id': gear['exchange'].get('rootfs-id', ''),
             'hash': 'v0-' + gear['exchange']['rootfs-hash'].replace(':', '-')
         })
-        signed_url = files.get_signed_url(file_path, file_system, filename='gear.tar', attachment=True, response_type='application/octet-stream')
+        signed_url = None
+        if file_system.is_signed_url():
+            #signed_url = files.get_signed_url(file_path, file_system, filename='gear.tar', attachment=True, response_type='application/octet-stream')
+            signed_url = file_system.get_signed_url(None, file_path)
         if signed_url:
             self.redirect(signed_url)
```
kofalt: Can this layer PR do anything more to remove signed-url conditionals from handler logic?

kofalt: My question is around the fact that this file is in ` jobs` module, and shouldn't know about the details of the underlying filesystem. I'm wondering if the work in this PR can reduce the cyclomatic complexity and abstraction violation of this `download` function by moving these conditionals outside this module.

Maybe a helper func that either starts writing the file or serves a redirect by passing the `RequestHandler` as a parameter (`self`)?

roycollins:
Per the review last Friday I have completed the changes that were requested.  I was also able to refactor out where the tempdir name was passed through some extra layers which simplified some of the logic and params.   

I did not change the logic on the get_signed_url but I can.  I will just need to dig a little deeper into the problem to fully understand the impacts.  

Please do a full review again.  There is still a little cleanup to polish up but I want to be sure the overall flow is complete first.  I will also rebase in the latest code and fix the conflict after we confirm everything looks good.   

Lastly I will squash my commits to cleanup the history before this branch finalized.

api/files.py
```diff
@@ -273,30 +261,30 @@ def get_file_path(file_info):
...
     still supports the legacy storage, attempt to serve from there.
     """
 
-    if config.fs.isfile(file_path):
-        return config.fs
-    elif config.support_legacy_fs and config.local_fs.isfile(file_path):
+    # For now this deprecated method will assume we are using a Py_FS type storage that has the isfile method.
+    # When we add more native storage types we will have to store the file system type in the file object and
+    # not rely on this method to determine where its physically located
+    if config.storage._fs.isfile(file_path):
```
ehlertjd: I think if we make `get_file_info` return Falsey (or raise) if the given file does not exist, then we can avoid going directly at `_fs` (or `get_fs()`)

api/handlers/listhandler.py
```diff
@@ -356,15 +356,24 @@ def _create_upload_ticket(self):
         if metadata is None or not filenames:
             self.abort(400, 'metadata and at least one filename are required')
 
-        tempdir = str(uuid.uuid4())
-        # Upload into a temp folder, so we will be able to cleanup
+        tempdir = None
```
ehlertjd: Remove this variable and just pass `None`?

api/jobs/handlers.py
ehlertjd: Agreed with @kofalt I think a helper function to serve a file is in order.

api/placer.py
```diff
@@ -20,12 +20,14 @@
 from .web.errors import FileFormException
 
 
+CHUNK_SIZE = 1024
```
ehlertjd: Use 1 MB chunk size

api/placer.py
```diff
@@ -583,20 +594,35 @@ def finalize(self):
...
+        # We could make this faster using a move if we know its a local to local fs move.
+        with config.local_fs.get_fs().open(tempZipPath, 'rb') as (f1
+                ), config.storage.open(token, util.path_from_uuid(token), 'wb', None) as f2:
+            while True:
+                data = f1.read(self._chunk_size)
+                if not data:
+                    break
+                f2.write(data)
+        # I belive these might be automatic inside the with statement
+        f1.close()
```
ehlertjd: Yep, these are covered by the `with` context, you can remove these lines.

roycollins:
I did one more pass of clean up and refactored out some more duplicated code.  Let's do one last review and if everything looks good Ill rebase in master. Otherwise, I think I will keep polishing this code for eternity.

api/data_views/file_opener.py
```diff
@@ -54,7 +54,7 @@ def __enter__(self):
         try:
             # Open the file using file_system
             file_path, file_system = files.get_valid_file(self.file_entry)
-            self._system_fd = file_system.open(file_path, open_mode)
+            self._system_fd = file_system.open(self.file_entry['_id'], file_path, open_mode, None)
```
ehlertjd: `_id` may not be set for older files - use `self.file_entry.get('_id')`

roycollins: good catch.  Fixed

api/upload.py
```diff
@@ -305,50 +342,33 @@ def clean_packfile_tokens(self):
...
         })
-
-        removed = result.deleted_count
+        removed = 0
+        for token in result:
+
+            try: 
+                config.local_fs.get_fs().removetree(token['_id'])
+            except:
+                self.log.error('Error removing token directory %s', token['_id'])
```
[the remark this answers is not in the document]
ehlertjd: This is mostly resolved, but now we have a potential disconnect between `cleaned` and `removed`. I.e. we can `remove` a token without `clean`ing a folder (it was already deleted)

api/files.py
```diff
@@ -177,22 +197,51 @@ def get_single_file_field_storage(file_system, use_filepath=False):
...
+            else:
+                self.filepath = util.path_from_uuid(self._uuid)
+
+            # Some placers reference path and others filepath so we use both to make it work for now
+            self.path = self.filepath
+
+            if not isinstance(self.filepath, unicode):
+                self.filepath = six.u(self.filepath)
+            
+            if type(file_system) is fs.tempfs.TempFS:
```
ehlertjd: I think this is always `False`, right?

roycollins: Yes, this was left over from the first iteration.  Removed.

api/upload.py
```diff
@@ -193,15 +214,22 @@ def _create_upload_ticket(self):
...
         signed_urls = {}
+        filedata = {}
         for filename in filenames:
-            signed_urls[filename] = files.get_signed_url(fs.path.join('tmp', tempdir, filename),
-                                                         config.fs,
-                                                         purpose='upload')
+            newUuid = str(uuid.uuid4())
+            signed_url = config.storage.get_signed_url(None, util.path_from_uuid(newUuid), purpose='upload'),
+            # TODO: this appears to come back as a tuple but if should be just a string
+            signed_urls[filename] = signed_url[0]
```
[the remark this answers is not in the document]
roycollins: This one I am not sure on. We will have to verify it on the live testing.  With the mocks it always seems to be a tuple.  Ill leave this open for now so we dont forget about it.

api/upload.py
```diff
@@ -130,15 +142,22 @@ def process_upload(request, strategy, access_logger, container_type=None, id_=No
...
-            raise Exception("file {} does not exist, files in tmpdir: {}".format(
-                field.path,
-                file_processor.temp_fs.listdir('/'),
-            ))
-        field.size = int(file_processor.temp_fs.getsize(field.path))
-        field.uuid = str(uuid.uuid4())
+
+        #We can trust the filepath on upload is accurate after form processing
+        if hasattr(field, 'filepath'):
+            filePath = field.filepath
```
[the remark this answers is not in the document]
roycollins: Correct, the filePath stuff ended up getting refactored out.  Its cleaned up now.

api/jobs/handlers.py
ehlertjd: Addendum: I think the utility function is out of scope for this PR, opening an additional PR to address it separately..

ehlertjd: Addressed in #1540

api/files.py
```diff
@@ -16,49 +16,37 @@
...
-            dst_path = six.u(dst_path)
-        dst_dir = fs.path.dirname(dst_path)
-        if not dst_fs:
-            dst_fs = self.persistent_fs
-        self._presistent_fs.makedirs(dst_dir, recreate=True)
-        if isinstance(self._temp_fs, fs.tempfs.TempFS):
-            fs.move.move_file(self._temp_fs, src_path, dst_fs, dst_path)
-        else:
-            self._presistent_fs.move(src_path=fs.path.join('tmp', self._tempdir_name, src_path), dst_path=dst_path)
+        newUuid = str(uuid.uuid4())
```
nagem: We should snake-case variables whenever possible.

api/handlers/listhandler.py
```diff
@@ -471,10 +477,16 @@ def get(self, cont_name, list_name, **kwargs):
...
-            signed_url = files.get_signed_url(file_path, file_system,
+            signed_url = None
+            if config.storage.is_signed_url():
+                try:
+                    signed_url = config.storage.get_signed_url(fileinfo.get('_id'), file_path,
                                               filename=filename,
                                               attachment=(not self.is_true('view')),
                                               response_type=str(fileinfo.get('mimetype', 'application/octet-stream')))
+                except fs.errors.ResourceNotFound:
+                    self.log.error('Error getting signed_url on non existing file')
```
nagem: Do we want to just log and error and move on in this instance? It's not clear from a high-level view what it would move on to in this case, but I might just be misreading.

api/upload.py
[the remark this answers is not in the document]
roycollins: cured

tests/integration_tests/python/test_download.py
```diff
@@ -213,7 +213,7 @@ def test_download_k(data_builder, file_form, as_admin, as_root, api_db, legacy_c
     # test missing file hangling
 
     file_id = api_db.acquisitions.find_one({'_id': ObjectId(acquisition)})['files'][0]['_id']
-    config.fs.remove(util.path_from_uuid(file_id))
```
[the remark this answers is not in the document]
roycollins: fixed

api/files.py
roycollins: yeah old habits.  There might be a few of those in this pull request. Fixed but Ill do a quick check for others.

api/handlers/listhandler.py
roycollins: I am not certain on this either.  I assume it because if you are trying to get parts of the file you wont have a signed url for the individual parts so then want to proceed through to the rest of the query params for Range and what not

ehlertjd: Looking at the structure of the code, we should not bury the exception here, if we get a ResourceNotFound, I believe we should be raising an `APINotFoundException`.

api/upload.py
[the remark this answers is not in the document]
ehlertjd: TIL: If you have a comma at the end of your statement (See Line 208) python will convert it into a tuple

api/handlers/listhandler.py
```diff
@@ -390,7 +396,7 @@ def build_zip_info(file_path, file_system):
         """
         Builds a json response containing member and comment info for a zipfile
         """
-        with file_system.open(file_path, 'rb') as f:
+        with file_system.open(None, file_path, 'rb', None) as f:
```
ehlertjd: The above function needs to be modified to take and pass a file_id.

api/handlers/listhandler.py
ehlertjd: This actually is never reached in testing - we bomb out on [line 440](https://github.com/flywheel-io/core/pull/1532/files#diff-2b6de8dffa100181ea2c9b85e6e5f50cR440) if the file doesn't exist.

I would still argue that we should just go ahead and raise an `APINotFoundException`

api/upload.py
```diff
@@ -130,28 +133,32 @@ def process_upload(request, strategy, access_logger, container_type=None, id_=No
...
+        #We can trust the filepath on upload is accurate after form processing
+        if hasattr(field, 'filepath'):
+            #Some placers need this value. Consistent object would be nice
+            field.path = field.filepath
+
+        # we can get the size from the file on disk but we need to know which fs its stored on.
+        if tempdir:
+            field.size = int(config.local_fs.get_fs().getsize(field.filepath))
+        else:
+            field.size = (config.storage.get_file_info(None, util.path_from_uuid(field.uuid)))['filesize']
```
ehlertjd: 1. Can't we use `get_file_info` in both cases now?
2. I think we have to pass the file uuid, right?

ehlertjd:
Getting close!

## Outstanding Items
- [ ] Raise `APINotFoundException` [on ResourceNotFound](https://github.com/flywheel-io/core/pull/1532#discussion_r257718795)
- [ ] Pass file `uuid` for [get_file_info](https://github.com/flywheel-io/core/pull/1532#discussion_r259123103)
- [ ] Packfile directories are not ending up in the `tokens` folder, and therefore not getting cleaned up if a packfile upload doesn't complete.

api/handlers/listhandler.py
roycollins: I found the specific case where returning a APINotFoundException is not working.  It happens when we have a storage provider that uses signedUrls but there are files in the storage provider using the legacy CAS file names.  The issue is the signed url can not be found since it assumes its using the uuid based filepaths and not the legacy CAS file paths.  It then fails over and just serves the file directly rather than using a signed URL.  If we can assure that no legacy CAS file structure is used on production instances then we can return the APINotFound exception.  But there is a specific integration test that checks this in test_downloads.py so I assume there must have been some customers where this check was needed.   I will leave the code as is for now.

api/storage/__init__.py
```diff
@@ -0,0 +1,14 @@
+"""Returns appropriate storage provider based on the url provided"""
+from .py_fs_storage import PyFsStorage
+
+def create_flywheel_fs(url):
+    """
+    This loads the storage provider based on the url provided
+    """
+    if url.startswith('gc::/'):
```
ehlertjd: Should be `gc://` right?

roycollins: Correct. This has already gotten abstracted out in the new changes with the packages.

api/storage/storage.py
```diff
@@ -0,0 +1,80 @@
...
+        """
+
+    @abstractmethod
+    def get_file_info(self, uuid, path_hint):
+        """
+        Returns basic file info about the referenced file object, None if the file does not exist
+
+        :param self: self reference
+        :param string uuid: internal fild uuid reference
+        :param path_hint string: internal reference to the file object on stroage, used when uuid is not available
```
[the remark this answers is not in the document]
roycollins: Storage interface has been addressed to support remove.  Testing it now.

PR #1533, opened 2019-02-08
[FLYW-408] Templates without session requirements 500

hkethi002:
### Review Checklist


api/dao/hierarchy.py
```diff
@@ -173,13 +173,13 @@ def check_session_for_single_template(session, template):
...
         a_requirements = template.get('acquisitions')
 
-        label = s_requirements.pop('label', s_requirements.pop('code', None))
-        if label:
-            m = re.match(label, session['label'])
-            if not m:
-                return False
-
         if s_requirements:
+            label = s_requirements.pop('label', s_requirements.pop('code', None))
```
nagem: Harsha is removing the `code` lookup since this is a session and also ensuring `label` exists just to be safe.

api/dao/hierarchy.py
```diff
@@ -173,13 +173,13 @@ def check_session_for_single_template(session, template):
...
+            if label:
+                m = re.match(label, session['label'])
```
nagem: >`label` exists just to be safe
here

swagger/schemas/definitions/project-template.json
```diff
@@ -20,7 +20,8 @@
                     "minItems": 1,
                     "items": {"$ref": "#/definitions/requirement"}
                 }
-            }
+            },
+            "additionalProperties": false
```
nagem: This means the FE fix @LinaJohnson is working on will need to go in at the same time. We might need to remove this if it is deemed necessary that we cut a release without it.

PR #1535, opened 2019-02-11
[FLYW-400] Add provider abstraction

ehlertjd:
This PR adds the [endpoints for the provider abstraction](https://flywheel-io.github.io/core/branches/compute-providers/swagger/index.html#/site/get_providers).

### Review Checklist


api/api.py
```diff
@@ -223,6 +223,13 @@ def prefix(path, routes):
 
         # Site
 
+        route('/site/providers',                ProviderHandler,   h='list',       m=['GET']),
```
hkethi002: Should this be `h='get_all'` to keep it consistent, or are we switching to list?

api/site/models/provider.py
```diff
@@ -0,0 +1,107 @@
...
+            'config': config
+        }
+
+    @property
+    def provider_id(self):
+        """Returns the unique ID of this document"""
+        return self._doc.get('_id')  # ID may not be set if this is a new provider
+
+    @provider_id.setter
+    def provider_id(self, _id):
```
hkethi002: What scenario would this be needed in? It seems like it would raise the ValueError most of the time.

api/api.py
ehlertjd: Probably?

api/site/models/provider.py
ehlertjd: This pattern is for one-time setting of the ID immediately after creation. Since we [insert](https://github.com/flywheel-io/core/pull/1535/files#diff-1336176fa866e5e9f3d2b0c9fe796ed3R15) a `Provider` object with an unset ID, the mapper [sets it based on inserted_id](https://github.com/flywheel-io/core/pull/1535/files#diff-1336176fa866e5e9f3d2b0c9fe796ed3R26). Otherwise it's an error to try to change a provider's ID.

We set on insert in-case the caller wants do anything else with that object (e.g. return the full object on insert) even though we're not doing anything with it in practice.

kofalt:
I'll take a look at this early next week hopefully, post rebase?

api/site/mappers/provider.py
```diff
@@ -0,0 +1,91 @@
+"""Provides data mapper for providers"""
+import datetime
+
+import bson
+
+from ... import config
+from .. import models
+
+class Provider(object):
```
hkethi002: I think mappers should be plural or have mapper in the name, I think it will make it more readable and easy to differentiate between the model.

PR #1538, opened 2019-02-14
[FLYW-425] Add ability to set providers on groups and projects

ehlertjd:
### Review Checklist


api/site/providers/repository.py
```diff
@@ -131,6 +133,74 @@ def update_provider(provider_id, doc):
...
+    3. A storage provider isn't already set
+
+    Setting either provider to the current value is a no-op and doesn't
+    trigger authorization errors.
+
+    Side-effect: This will convert any IDs in the links parameter to ObjectIds.
+
+    Args:
+        container (dict): The current container (or empty if it doesn't exist)
+        links (dict): The provider links to update.
```
hkethi002: I find links to be a confusing word to use here, can't we just call them provider_ids

ehlertjd: Agreed, I changed the variable & function names.

api/site/providers/repository.py
```diff
@@ -131,6 +133,74 @@ def update_provider(provider_id, doc):
...
+        is_admin (bool): Whether or not the user is a site administrator
+
+    Raises:
+        APIPermissionException: If the user is unautorized to make the change.
+        APIValidationException: If the user attempted an invalid transition.
+        APINotFoundException: If the given storage provider does not exist.
```
nagem: I like the additional details in the function signature, especially the `Raises` 👍

api/site/providers/repository.py
```diff
@@ -131,6 +133,66 @@ def update_provider(provider_id, doc):
...
+
+    for provider_class in ('compute', 'storage'):
+        updates[provider_class] = False
+        if provider_class in provider_ids:
+            # Ensure ObjectId
+            provider_ids[provider_class] = bson.ObjectId(provider_ids[provider_class])
+            current_id = current_provider_ids.get(provider_class)
+            if current_id != provider_ids[provider_class]:
+                if current_id:
+                    raise errors.APIValidationException('Cannot change {} provider once set!'.format(provider_class))
```
nagem: If a site admin sets a group provider, does that mean the project provider remains empty and is collected from the group level until one at the project level is set?

ehlertjd: @nagem That's a good question. Thinking about it, I think we do have to propagate down from the group level (setting on any project that hasn't already been set.)

If we start using implicit provider selection, then we could end up with data split across projects if you were ever to finally set a different provider on a project that had implicitly been using the group provider.

ehlertjd: As discussed offline, we will not propagate changes, and will look up to find the provider.

PR #1540, opened 2019-02-15
Add utility function to generically serve a file

ehlertjd:
Addresses the comment [here](https://github.com/flywheel-io/core/pull/1532#discussion_r256025303)

### Review Checklist


api/jobs/handlers.py
```diff
@@ -321,19 +321,13 @@ def download(self, **kwargs): # pragma: no cover
         """Download gear tarball file"""
         dl_id = kwargs.pop('cid')
         gear = get_gear(dl_id)
-        file_path, file_system = files.get_valid_file({
-            '_id': gear['exchange'].get('rootfs-id', ''),
+
+        file_id = gear['exchange'].get('rootfs-id', '')
+        file_path, storage = files.get_valid_file({
+            '_id': file_id,
             'hash': 'v0-' + gear['exchange']['rootfs-hash'].replace(':', '-')
```
kofalt: Checking my understanding here:

1. Does one of (`file_id`, `file_path`) have to be non-empty?
1. Why default `file_id` to empty string instead of None?
1. What is the behavior of `send_or_redirect_file` and in turn `storage.get_signed_url` / `set_for_download` if both or neither keys are valid?

ehlertjd: All good questions - Originally I did the refactor without changing any behavior, but I can update the function signature.

1. Yes
2. 🤷‍♀️ Probably should be `None`? I didn't change any _behavior_ as part of this refactor.
3. Opening a non-existent file will result in a `raise`

ehlertjd:
@kofalt comments answered / addressed

kofalt:
Enjoy your rebase

PR #1541, opened 2019-02-19
[FLYW-395] Compute provider Google Cloud

roycollins:
Creates the Google Cloud Compute Provider 

We need to finalize the GCP permissions list but we can adjust that as needed. 

### Review Checklist


api/site/providers/gc_compute_provider.py
```diff
@@ -0,0 +1,312 @@
...
+            if r.get('permissions') and len(r['permissions']) != len(small_perms):
+                missing += list(set(small_perms) - set(r['permissions']))
+            i += 1
+
+        if missing:
+            print 'We are missing {} permissions out of {}'.format(len(missing), len(PERMS))
+            raise errors.APIValidationException('Your account is missing the following permissions' + ' '.join(missing))
+
+
+PERMS = [
```
nagem: Can we just do `compute.*` here?

roycollins: Unfortunately they explicitly disallow allow wild cards in the permissions list.

This is going to be a larger issue when we do the inverse check as we will have to update the permission list when google adds new permissions, and we will never know when that is.  I am still looking to see if they have an API call that we can make to get all the permissions but so far I have not found anything.

nagem: I'm worried about that situation too - Google updates something and no one can upload new Compute Providers until we cut a release and deploy. Maybe we split the endpoints - one to check, one to save? The check could give them information via the UI about how you might have over or under provisioned the service account. The endpoint to save only checks that the required permissions are there.

api/site/providers/gc_compute_provider.py
```diff
@@ -0,0 +1,312 @@
...
+import requests
+import jwt
+
+from ...web import errors
+from .compute_provider import ComputeProvider
+from .factory import ProviderKey
+from ..models import ProviderClass
+from ...config import log
+
+# pylint: disable=too-few-public-methods
```
kofalt: That's real silly - just disable it globally.

api/site/providers/gc_compute_provider.py
roycollins: For sure,  We will only stop adding a new provider if you don't hit the minimum requirements.  The other half of the check can be only for UI. That part is still yet to be built. Its just the inverse of the logic but it does not have a home at the moment.  We can flesh out the details once we bring the frontend and backend together.

nagem: Ah thanks I didn't read closely enough to distinguish the difference.

nagem:
screen shot 2019-02-20 at 3 39 01 pm
I'm getting really weird errors trying to add comments to this PR, anyone else having issues?

api/site/providers/gc_compute_provider.py
```diff
@@ -0,0 +1,312 @@
...
+    'compute.disks.delete',
+    'compute.disks.get',
+    'compute.disks.getIamPolicy',
+    'compute.disks.list',
+    'compute.disks.resize',
+    'compute.disks.setIamPolicy',
+    'compute.disks.setLabels',
+    'compute.disks.update',
+    'compute.disks.use',
+    #'compute.disks.*',
```
mjburling: 👍 I like this method of in-line documentation. If only wildcards were accepted.

swagger/paths/providers.yaml
```diff
@@ -0,0 +1,89 @@
+/site/providers:
+  get:
+    summary: Return a list of all providers on the site
+    operationId: get_providers
+    tags:
+    - site
```
mjburling: I'd love to see some consistency in indentation depth for the yaml here and elsewhere. I'd settle for single-line json-style over inconsistent multi-line arrays:
```
acceptable_json_array: ['singel_value']
good_yaml_array:
  - single_value
questionable_yaml_array:
- single_value
```

You'd think this would already be linted away, though.

api/site/providers/gc_compute_provider.py
mjburling: I second this. I've actually never found this particular stumbling block useful.

If it were me, I'd add `too-few-public-methods` to the end of [this line in .pylintrc](https://github.com/flywheel-io/core/blob/compute-providers-gc/tests/.pylintrc#L63) and call it a day.

api/site/mappers/provider_mapper.py
```diff
@@ -0,0 +1,91 @@
+"""Provides data mapper for providers"""
```
ehlertjd: I don't think this file should be here. Probably left over from a merged rename?

api/site/mappers/site_settings_mapper.py
```diff
@@ -0,0 +1,54 @@
+"""Provides data mapper for site settings"""
```
ehlertjd: I don't think this file should be here either

api/site/providers/gc_compute_provider.py
```diff
@@ -0,0 +1,318 @@
...
+
+    def validate_config(self):
+        """
+            Confirms the minimum required config is set
+        """
+
+        if not self.config:
+            raise errors.APIValidationException('Google Compute configuration is required')
+
+        if not self.config.get('client_id'):
```
ehlertjd: If we don't want to rely on JSON schemas, could line 37-53 just be a for loop?
```python
for key in ('client_id', 'client_email', ...):
  if not self.config.get(key):
    raise errors.APIValidationException('Google Compute requires {} be set'.format(key))
```

api/site/providers/gc_compute_provider.py
```diff
@@ -0,0 +1,318 @@
...
+        if not self.config.get('client_x509_cert_url'):
+            raise errors.APIValidationException('Google Compute requires client_x509_cert_url be set')
+
+        if not self.config.get('project_id'):
+            raise errors.APIValidationException('Google Compute requires project_id be set')
+
+        self._validate_permissions()
+
+    def get_redacted_config(self):
+        return {
```
ehlertjd: From the [super class](https://github.com/flywheel-io/core/blob/master/api/site/providers/base.py#L36):
> Return the configuration with all private key material and credentials replaced with None.

Let's include private keys but with values of `None` so that the UI has a hint that those fields exist but are private.

api/site/providers/gc_compute_provider.py
```diff
@@ -0,0 +1,318 @@
...
+
+            if not small_perms:
+                break
+
+            payload = {
+                'permissions': small_perms
+            }
+
+            # Using cloud resource manager we can check the permissions for the current user without any permissions
+            # But the Cloud Resource Manager API must be enabled in the project
```
ehlertjd: Is this enabled by default? We'll need to document if not.

api/site/providers/gc_compute_provider.py
```diff
@@ -0,0 +1,318 @@
...
+            response = requests.post('https://cloudresourcemanager.googleapis.com/v1/projects/{}:testIamPermissions/' \
+                .format(self.config['project_id']),
+                                     headers=headers,
+                                     data=payload)
+
+            if response.status_code == 403:
+                error = response.json()
+                if 'Cloud Resource Manager API has not been used' in error['error']['message']:
+                    raise errors.APIValidationException("You need to enable the Clour Resource Manager API for your project")
```
ehlertjd: Clour->Cloud

api/site/providers/gc_compute_provider.py
```diff
@@ -0,0 +1,318 @@
...
+                else: # All other permission errors but the API has been enabled
+                    log.error(error['error']['message'])
+                    # TODO test to see what else causes 403 errors on this API and filter out is possible
+                    raise errors.APIValidationException(error['error']['message'])
+
+            if response.status_code == 404:
+                error = response.json()
```
ehlertjd: Do we use `error`, or can this line be removed?

api/site/providers/gc_compute_provider.py
```diff
@@ -0,0 +1,318 @@
...
+                log.error(response.raw)
+                raise errors.APIValidationException('Unable to validate permissions. Be sure the procject_id is correct')
```
ehlertjd: procject_id -> project_id

api/site/providers/gc_compute_provider.py
ehlertjd: @mjburling @kofalt Does this set of permissions look good? Or can it be culled a bit?

api/site/providers/gc_compute_provider.py
```diff
@@ -0,0 +1,318 @@
...
+
+        if missing:
+            print 'We are missing {} permissions out of {}'.format(len(missing), len(PERMS))
+            raise errors.APIValidationException('Your account is missing the following permissions' + ' '.join(missing))
+
+
+PERMS = [
+    'compute.acceleratorTypes.get',
+    'compute.acceleratorTypes.list',
+    'compute.addresses.get',
```
ehlertjd: Can you add a comment above that tells how to get / update this list?

api/site/providers/gc_compute_provider.py
```diff
@@ -0,0 +1,318 @@
...
+        self._validate_permissions()
+
+    def get_redacted_config(self):
+        return {
+            'client_email': self.config['client_email'],
+            'client_id': self.config['client_id'],
+            'project_id': self.config['project_id']
+        }
+
+    def _validate_permissions(self):
```
ehlertjd: How can we test this function? Any way we could include it in CI? I think it's worth looking into.

roycollins:
Changes were made as requested.  

Whats left is to document the setup process.  If there is an Ops setup manual we should add a step to details how to turn on the APIs for Google Cloud.  

As for testing Google does not offer a test account for validating API calls.  Others have fully mocked the responses to generate matching hash pairs. Some just manually copy in valid Bearer Tokens. But really we should have it tested in our end to end tests where we use a valid set of dev only test credentials that perform all tasks from account setup to creating instances on the dev GCP account.

For now, the error will specifically alert if you forget to enable/config something. So hopefully the error will be enough info for the Ops group to determine what is wrong when they are enabling GC compute but it might not be enough if we allow customers to enable it themselves.

ehlertjd:
@roycollins I think if this moves over to the storage-plugins repo, we may some expanded options for testing, as the CI process is configured with a set of valid credentials for GCP/AWS

ehlertjd:
This landed in a separate repo

PR #1542, opened 2019-02-20
Update URLFileWrapper to read directly from response

ehlertjd:
BufferedReader was choking in the actual signed-url case because the HTTPReponse is closed when everything is read. I'm not convinced that BufferedReader was providing any gains, so we're reading directly from the response instead.

This fixes the root cause of CI failures further down the line.

### Review Checklist


api/io/url_file_wrapper.py
```diff
@@ -17,7 +13,6 @@ def __init__(self, url, http):
         self.http = http
 
         self._response = None
-        self._reader = None
 
     def __enter__(self):
         self._response = self.http.request('GET', self.url, preload_content=False)
```
hkethi002: I don't know if I'm understanding this correctly but do we need to set `release_conn=False` in the arguments so that we keep the connection until we finish reading the response, or does the response persist even if the connection is returned?

ehlertjd: Following the guideline [here](https://urllib3.readthedocs.io/en/latest/advanced-usage.html#streaming-and-io):
> When using preload_content=False, you should call release_conn() to release the http connection back to the connection pool so that it can be re-used.

PR #1544, opened 2019-02-20
Allow users with project permissions to resolve groups [FLYW-612]

hkethi002:
Users should be able to resolve any subset of a path they can already resolve.

#### Example:
If `fw ls scitran/Neuroscience` returns something other than a 403 for me, I should be able to see Neuroscience with `fw ls scitran` but only the projects I have access to.
And with `fw ls` I should see scitran even if I don't have permission for the group but I do for a child.

Not sure if there's a better way to do the `get_resolvable_groups` function, or if we should just implement a get_projects_groups endpoint which may help frontend out too.
### Review Checklist

### DIFF
--- a/api/dao/containerutil.py
+++ b/api/dao/containerutil.py
@@ -140,7 +140,7 @@ def extract_subject(session, project):
     return subject
 
 
-def get_resolvable_groups(uid):
+def get_project_groups(uid):
     pipeline = [
         {
             '$match': {'permissions._id': uid}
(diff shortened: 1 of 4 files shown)


hkethi002:
Found out the get groups from projects code already exists, going to try using that instead

api/dao/containerutil.py
```diff
@@ -140,6 +140,20 @@ def extract_subject(session, project):
     return subject
 
 
+def get_resolvable_groups(uid):
```
ehlertjd: Add a doc string please.

PR #1545, opened 2019-02-25
[FLYW-481] Test that a user with access to the project can download analyses

hkethi002:
### Review Checklist


api/data_export/strategy/hierarchy.py
```diff
@@ -109,6 +108,13 @@ def identify_targets(self, spec, uid, summary):
             else:
                 parents = self._resolve_parents(container)
 
+            if container_type == 'analysis' and uid:
+                if not summary:
```
ehlertjd: I think we need to drop this if statement, right @hkethi002?

hkethi002: Yep, I'll do that

PR #1547, opened 2019-02-28
Master subject code

davidfarkas:
Add handler which implements the master subject code workflow's first step.

- Send patient ID (e.g., MRN), first name, last name, DOB
- Indicate whether to use patient ID for identification or first name, last name, DOB
- Generated codes should be non-consecutive, 6-character uppercase alphanumeric strings [0-9A-Z], prefixed by fw, e.g., fw-ABCD12

See also: https://docs.google.com/document/d/1nz93AKrVRl4D9Bz_DuXBSgv1FAzOF_uR2y5rSjvad2g/edit#heading=h.qso2e6mxc4sh

### Review Checklist


ehlertjd:
I haven't reviewed yet, but historically I've seen characters omitted from a pool when generating random strings for two reasons:
1. Prevent look-alike symbols (e.g. 0 vs O and I vs 1)
2. Prevent accidentally generating expletives.

I would suggest removing zero and all vowels from the character pool when generating subject codes.

api/config.py
```diff
@@ -48,7 +48,10 @@
         'registered': False,
         'ssl_cert': None,
         'inactivity_timeout': None,
-        'upload_maximum_bytes': '10737418240'
+        'upload_maximum_bytes': '10737418240',
+        'master_subject_code_size': 6,
```
hkethi002: The default here should be a string, or the config value should be casted to an int before being applied because Env Var values are always strings.

The error you get when POST with a configured value for the master_subject_code_size:
`"message": "range() integer end argument expected, got unicode."`

swagger/schemas/definitions/master-subject-code.json
```diff
@@ -0,0 +1,83 @@
...
+      "description": "First name"
+    },
+    "lastname": {
+      "type": "string",
+      "maxLength": 64,
+      "description": "Last name"
+    },
+    "date_of_birth": {
+      "type": "string",
+      "description": "Date of birth"
```
hkethi002: Do we want to enforce a format for the DOB? Require it to be `YYYY-MM-DD`?

tests/unit_tests/python/test_master_subject_code.py
```diff
@@ -0,0 +1,53 @@
...
+from api.handlers.mastersubjectcodehandler import generate_master_subject_code
+
+
+def test_master_subject_code_handler(as_user, mocker, config, api_db):
+    # if the generated subject code already exists,
+    # it will try to generate new ones for 5 seconds,
+    # but after that it gives up (Birth day problem)
+    # In this case we can increase the length of the code
+    mock_gen_code = mocker.patch('api.handlers.mastersubjectcodehandler.generate_master_subject_code')
+    mock_gen_code.return_value = 'fw-ABC123'
```
hkethi002: Another minor nitpick, but this value isn't valid with the default configuration since we don't allow the character A in the code

api/api.py
```diff
@@ -369,6 +370,7 @@ def prefix(path, routes):
         route('/<par_cont_name:groups>/<par_id:{gid}>/<cont_name:projects>', ContainerHandler, h='get_all', m=['GET']),
         route('/<par_cont_name:{cname}>/<par_id:{oid}>/<cont_name:{cname}>', ContainerHandler, h='get_all', m=['GET']),
 
-
+        # Master Subject Code
+        route('/master-subj-code', MasterSubjectCodeHandler, m=['POST']),
```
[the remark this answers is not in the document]
ehlertjd: I would prefer `/subjects/master-code` even more.

api/handlers/mastersubjectcodehandler.py
```diff
@@ -0,0 +1,71 @@
...
+        payload = self.request.json_body
+
+        validate_data(payload, 'master-subject-code.json', 'input', 'POST')
+
+        if payload['use_patient_id']:
+            query = {
+                'patient_id': payload['patient_id']
+            }
+        else:
+            query = {
```
ehlertjd: Should we normalize these values? (i.e. trim leading/trailing whitespace, force to lower case) Or do we want "Robert Smith" to be different from "robert smith"?

api/handlers/mastersubjectcodehandler.py
```diff
@@ -0,0 +1,71 @@
...
+                    chars=config.get_item('site', 'master_subject_code_chars'),
+                    prefix=config.get_item('site', 'master_subject_code_prefix')
+                )
+                try:
+                    result = config.db.master_subject_codes.insert_one(payload)
+                    break
+                except errors.DuplicateKeyError:
+                    # Stop generating new codes after 5 seconds to avoid timeout
+                    if time.time() - start > 5:
+                        return self.abort(500, "Couldn't generate new master subject code.")
```
ehlertjd: If this starts happening we should be logging a critical error as well so that CE can monitor and react.

In general we should try to avoid `abort` calls, raise an [Error](https://github.com/flywheel-io/core/blob/master/api/web/errors.py) instead.

api/dao/containerutil.py
```diff
@@ -125,6 +125,11 @@ def extract_subject(session, project):
         # If a subject with that id exists and is deleted, create a new one
         if result:
             subject['_id'] = bson.ObjectId()
+    elif subject.get('master_code'):
```
ehlertjd: Does this effectively allow me to create two subjects with the same `code` but different `master_code` in one project? I could see that being potentially confusing if it ever happened.

Is there any requirement that the specified `master_code` exist?

api/handlers/mastersubjectcodehandler.py
```diff
@@ -0,0 +1,71 @@
...
+                    size=config.get_item('site', 'master_subject_code_size'),
+                    chars=config.get_item('site', 'master_subject_code_chars'),
+                    prefix=config.get_item('site', 'master_subject_code_prefix')
+                )
+                try:
+                    result = config.db.master_subject_codes.insert_one(payload)
+                    break
+                except errors.DuplicateKeyError:
+                    # Stop generating new codes after 5 seconds to avoid timeout
+                    if time.time() - start > 5:
```
ehlertjd: Maybe make this timeout a constant so that we can use a different value in the unit test and not have to wait 5 seconds?

swagger/schemas/definitions/master-subject-code.json
davidfarkas: Currently we are using `jsonschema` `2.6.0` which supports only JSON schema draft 3 and draft 4. `date` format was added in draft 7. `jsonschema` `3.0.x` supports that. Do we want to upgrade or do the validation in the handler?

see: https://json-schema.org/understanding-json-schema/reference/string.html
https://pypi.org/project/jsonschema/2.6.0/
https://pypi.org/project/jsonschema/

ehlertjd: In lieu of schema support, I think that yes we should document the format, and enforce it in the handler.

api/handlers/mastersubjectcodehandler.py
```diff
@@ -0,0 +1,71 @@
...
+                'first_name': payload['first_name'],
+                'last_name': payload['last_name'],
+                'date_of_birth': payload['date_of_birth']
+            }
+
+        payload.pop('use_patient_id')
+        subj_codes = list(config.db.master_subject_codes.find(query))
+
+        if len(subj_codes) > 1:
+            self.abort(400, "Found multiple master subject codes")
```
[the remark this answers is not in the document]
davidfarkas: The idea was here to allow inserting two (or more) patient with the same name and date of birth, but with different patient id. I think it is a valid use case at a big site. Suppose that A and B are these two patient.

- first time, I can create a master subject code for A via patient id or name and DOB
- at this point, if I want to create a new code for patient B I have to use its patient id otherwise if patient A was inserted with name and DOB info I will get A's master code
- after I inserted A and B I will have to use their patient id to uniquely identify them. (Yep, I should improve the error message)

We wouldn't want to violate the response schema by returning a list when we found multiple matches, but we also want to let inserting multiple patient with the same name and DOB.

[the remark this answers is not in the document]
ehlertjd: @davidfarkas That use case makes sense, and I agree that the error message could be clearer.

However, I'm concerned about there being a race condition where if two people post the same First/Last/DOB or PatientID at roughly the same time you could end up with two master subject codes for one set of key material.

Given that we expect [5 seconds to be the worst case scenario](https://github.com/flywheel-io/core/pull/1547/files#diff-f919b860372e2e8ecc9ad2563b25a269R68)) between checking if a subject exists and inserting it, I could see this happening for tools that do multi-processing.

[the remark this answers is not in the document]
davidfarkas: Good point. According to this conversation I updated the code. Now I search for existing master code in the loop, so it will generate the new code almost immediately. I added a unique compound index on first name, last name, date of birth and patient id, and other unique index on patient id. I think these changes solve all the above mentioned problems.

api/dao/containerutil.py
davidfarkas: I updated the placers to validate the master code through an endpoint if provided. See the changes in `placer.py`.

davidfarkas: Yep, that's right.

I missed that part of your question. What do you think adding two unique compound index, one on `code` and `project` and another on `master_code` and `project`? I think that solves the problem.

api/placer.py
```diff
@@ -89,6 +90,19 @@ def requireMetadata(self):
         if self.metadata == None:
             raise FileFormException('Metadata required')
 
+    def verify_master_subject_code(self):
+        if self.metadata and self.metadata.get('session', {}).get('subject', {}).get('master_code'):
+            url = config.get_item('site', 'master_subject_code_verify_url')
+            if url:
+                url = url.rstrip('/')
+                resp = requests.get(url + '/' + self.metadata['session']['subject']['master_code'])
```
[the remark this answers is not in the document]
davidfarkas: If you think that we can just skip supporting third party msc service for now then I agree we should do that.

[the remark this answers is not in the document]
davidfarkas: hm, if we remove this then I think we will support only the first scenario from the docs: https://docs.google.com/document/d/1nz93AKrVRl4D9Bz_DuXBSgv1FAzOF_uR2y5rSjvad2g/edit#heading=h.fs7cjn4gwaf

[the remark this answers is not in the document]
ehlertjd: What I was getting at is:
I would much prefer that we use a local function to check the database if the `master_subject_code_verify_url` is not set (or effectively localhost). I'm not positive that the above approach would work in all deployments.

api/handlers/mastersubjectcodehandler.py
```diff
@@ -0,0 +1,99 @@
...
+                'patient_id': payload['patient_id']
+            }
+        else:
+            query = {
+                'first_name': payload['first_name'],
+                'last_name': payload['last_name'],
+                'date_of_birth': payload['date_of_birth']
+            }
+
+        payload.pop('use_patient_id')
```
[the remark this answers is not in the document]
davidfarkas: Updated the code to use model/handler/mapper approach.

api/handlers/mastersubjectcodehandler.py
```diff
@@ -19,12 +23,30 @@ def generate_master_subject_code(size=6, chars=string.ascii_uppercase + string.d
 
 
 class MasterSubjectCodeHandler(base.RequestHandler):
-    @require_login
+
+    def verify_code(self, _id):
+        master_code = config.db.master_subject_codes.find_one({'_id': _id})
+        if not master_code:
+            raise errors.APINotFoundException("This master subject code doesn't exists")
+        return {}
```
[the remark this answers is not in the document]
davidfarkas: dropped the return, added `@require_login`

tests/integration_tests/python/test_master_subject_code.py
```diff
@@ -1,88 +1,123 @@
 import re
 
-def test_master_subject_code(as_public, as_user, api_db):
+import pytest
+from pymongo import errors
+
+
+def test_master_subject_code(as_public, as_admin, api_db):
```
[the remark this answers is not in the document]
davidfarkas: It was tested implicitly, but now I added a dedicated test for it.

api/dao/containerutil.py
davidfarkas: Updated other places that use `extract_subject`. Also added unique compound indexes to solve the same code different master code problem.

api/master_subject_code/models/master_subject_code.py
```diff
@@ -0,0 +1,37 @@
...
+        Arguments:
+            **kwargs -- Fields to set
+        """
+
+        for key in kwargs.keys():
+            if self.get(key) is None:
+                self[key] = kwargs[key]
+
+    @staticmethod
+    def generate_code(length, allowed_chars, prefix):
```
hkethi002: I'm not sure if this function needs to be here, could we just call the underlying generate_code directly from the handler?

api/config.py
```diff
@@ -42,7 +42,11 @@
         'registered': False,
         'ssl_cert': None,
         'inactivity_timeout': None,
-        'upload_maximum_bytes': '10737418240'
+        'upload_maximum_bytes': '10737418240',
+        'master_subject_code_size': '6',
+        'master_subject_code_prefix': 'fw',
+        'master_subject_code_chars': 'BCDFGHJKLMNPQRSTVWXYZ123456789',
+        'master_subject_code_verify_config': None
```
nagem: This increases the number of site config values by 50% - do we want to consider subclassing them?

api/master_subject_code/handlers/master_subject_code_handler.py
```diff
@@ -0,0 +1,89 @@
...
+        if not master_code:
+            raise errors.APINotFoundException("This master subject code doesn't exists")
+
+    @require_admin
+    def post(self):
+        """Generate or find master subject code for the given patient."""
+        payload = self.request.json_body
+        validate_data(payload, 'master-subject-code.json', 'input', 'POST')
+        use_patient_id = payload.pop('use_patient_id')
+        master_subject_codes = mappers.MasterSubjectCodes()
```
ehlertjd: Could use `self.mapper` here?

bin/database.py
```diff
@@ -2330,6 +2330,19 @@ def upgrade_to_63():
...
     process_cursor(cursor, upgrade_template_to_list)
 
+
+def upgrade_to_64():
+    '''
+    (project, code) compound index should be unique in subjects collection
+    '''
+    # cleanup codes with empty and null value
+    config.db.subjects.update_many({'code': {'$in': [None, '']}}, {'$unset': {'code': 1}})
+    # apply indexes
```
ehlertjd: By virtue of pulling in config, aren't we going to implicitly create these indexes? What happens in that scenario?

davidfarkas: The `initialize_db` runs only when one of the `get_item`, `get_auth`, `get_feature`, `get_config`, `get_public_config` method is called from the config module. Actually the `get_config` method calls `initialize_db`. So we create the indexes lazily. I verified that we don't call any of these methods from `database.py`. It's not too explicit, maybe later we could make it more implicit. Currently the `app_factory` method also takes advantage of the lazy db init (`web/start.py:97`)

api/master_subject_code/handlers/master_subject_code_handler.py
davidfarkas: Thanks, I missed that.

api/master_subject_code/models/master_subject_code.py
davidfarkas: Sure, initially there was a factory method in my mind, but I agree that in the current form it doesn't have any sense.

api/config.py
davidfarkas: I put all of these under a new `master_subject_code` key. What do you think now?

sample.config
```diff
@@ -22,6 +22,10 @@
 #SCITRAN_SITE_REGISTERED=""
 #SCITRAN_SITE_SSL_CERT=""
 #SCITRAN_SITE_UPLOAD_MAXIMUM_BYTES=10737418240		# Unit is Bytes
+#SCITRAN_SITE_MASTER_SUBJECT_CODE_SIZE=6
+#SCITRAN_SITE_MASTER_SUBJECT_CODE_PREFIX="fw"
+#SCITRAN_SITE_MASTER_SUBJECT_CODE_CHARS="BCDFGHJKLMNPQRSTVWXYZ123456789"
+#SCITRAN_SITE_MASTER_SUBJECT_CODE_VERIFY_URL=""
```
hkethi002: I think these would need to change to just `SCITRAN_MASTER_SUBJECT_CODE`

PR #1549, opened 2019-03-04
User info, jobs

davidfarkas:
Originally these changes were created for the GHC integration demo. 
We store the GCP import profiles on user's `info` field. The `/users/self/jobs` endpoint was useful to list the import jobs on GHC frontend.

### Review Checklist


api/handlers/userhandler.py
```diff
@@ -204,3 +204,46 @@ def _get_user(self, _id):
...
+        result = self.storage.modify_info(self.uid, payload)
+        return {'modified': result.modified_count}
+
+    @require_login
+    def get_jobs(self):
+        query = {'origin.id': self.uid}
+        if self.get_param('gear', None):
+            query['gear_info.name'] = self.get_param('gear')
+
+        jobs = config.db.jobs.find(query, sort=[('created', -1)])
```
ehlertjd: We should probably support pagination on this endpoint, the list of jobs for an individual user can be quite large.

api/handlers/userhandler.py
```diff
@@ -204,3 +204,46 @@ def _get_user(self, _id):
...
+            'stats': {
+                'complete': 0,
+                'pending': 0,
+                'failed': 0,
+                'running': 0,
+            },
+            'jobs': []
+        }
+
+        for job in jobs:
```
ehlertjd: Any way we could leverage the [job Queue logic](https://github.com/flywheel-io/core/blob/master/api/jobs/queue.py#L705)?

api/handlers/userhandler.py
davidfarkas: Updated the code, now I use `dbutil.paginate_find`.

api/handlers/userhandler.py
```diff
@@ -204,3 +206,39 @@ def _get_user(self, _id):
...
+        return {'modified': result.modified_count}
+
+    @require_login
+    def get_jobs(self):
+        whitelist = {'created-by': [self.uid]}
+        if self.get_param('gear', None):
+            whitelist['gear-name'] = [self.get_param('gear')]
+
+        query = Queue.lists_to_query(whitelist, {}, [])
+        page = dbutil.paginate_find(config.db.jobs, {'filter': query}, self.pagination)
```
ehlertjd: Do we need to log access here? 
Also, this will return jobs with `_id` field, whereas the schema says they'll have an `id` field.

davidfarkas:
> Please add access checks and make changes (if necessary) for the following assertions:
> 
> * User `info` should NOT come back on list users request
> * User `info` should NOT come back on GET request EXCEPT if the requested user is the logged in user.

Do we want the same behavior in case of admin user?

ehlertjd:
@davidfarkas Is my understanding that we were using this field to store some cloud credentials? If that's the case, then I don't think even admin users should be able to see the info field here.
@nagem thoughts?

davidfarkas:
@ehlertjd we will store only connection profiles on user's info field e.g. google project name, healthcare dataset name, dicom, fhir, and hl7 data store names. We will store the cloud credentials in the `authtokens` collection where currently the login credentials are.

ehlertjd:
@davidfarkas If that's the case, then I'd amend the assertions as follows:
* User info should NOT come back on list users request
* User info should NOT come back on GET request EXCEPT if the requested user is the logged in user OR a site administrator.

nagem:
Just catching up but 👍from me on final decision.

PR #1550, opened 2019-03-04
Fal storage plugins

roycollins:
Convert abstraction layer to use the new plugins module from flywheel-common.

### Review Checklist


api/config.py
```diff
@@ -319,10 +320,9 @@ def get_release_version():
     return release_version
 
 # Storage configuration
-
-storage = create_flywheel_fs(__config['persistent']['fs_url'])
+PRIMARY_STORAGE = storage.create_flywheel_fs(__config['persistent']['fs_url'])
```
ehlertjd: Can we keep this lowercase so that it's consistent with the other globals in this module?
Or even better, we could have done `from flywheel_common.storage import create_flywheel_fs` and kept the same variable name (and have minimal changes in the diff)

api/util.py
```diff
@@ -267,6 +267,9 @@ def send_or_redirect_file(handler, storage, file_id, file_path, filename,
 
 def format_hash(hash_alg, hash_):
     """
+    @deprecated
+    use the version in flywheel common
+
```
ehlertjd: Let's just delete them and update any usages.

api/config.py
roycollins: I converted it to lower case but I kept the primary_storage name since we have local_fs and legacy_fs.   Storage is the interface and the local, primary, and legacy are the instantiations of the module.  I could rename local_fs to local_storage for consistency since its not a file system anymore, but rather a storage abstraction. Let me know your thoughts on this one.

api/util.py
roycollins: Done.

api/config.py
ehlertjd: I think consistency is good, but probably out of scope for this PR. Let's address when we formalize storage providers, and leave it for now, unless you really want to make the change.

api/config.py
```diff
@@ -319,10 +320,9 @@ def get_release_version():
     return release_version
 
 # Storage configuration
-
-storage = create_flywheel_fs(__config['persistent']['fs_url'])
+primary_storage = storage.create_flywheel_fs(__config['persistent']['fs_url'])
```
nagem: If `storage` is now `primary_storage` is that the "center's" storage in a multi-project sense? What would an example of "non-primary" storage be?

roycollins: Primary Storage is the centers persistent_fs storage. Im not sure how that varies on multi-project but I assume they have only one persistent_url for the center.  

An example of non-primary storage would be legacy storage that is still configured or the local storage that is configured on the flywheel install itself; this might be a NAS block or local disks on the flywheel infrastructure, whereas the persistent,  primary_storage, might be a different NAS, local disk, or object storage location.  

Primary storage was originally 'fs'.  But since it's not a file system anymore it was renamed to primary_storage.

ehlertjd:
Just needs a rebase and LGTM

api/data_export/strategy/classic.py
```diff
@@ -75,8 +75,7 @@ def visit_file(self, parents, parent_type, file_group, file_entry, summary):
 
             return [
                 models.DownloadTarget('file', dst_path, parent_type, parent_id, file_entry['modified'],
-                    file_entry['size'], file_entry.get('type'), file_id=file_entry.get('_id'),
-                    filename=file_entry['name'], file_group=file_group, src_path=src_path)
+                    file_entry['size'], file_entry.get('type'), file_id=file_entry.get('_id'), src_path=src_path)
```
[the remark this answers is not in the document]
roycollins: Good catch. I went ahead and fixed that.  Is there a place where I can add a test case to catch this in the future?

PR #1552, opened 2019-03-04
Storage plugin fixes

ehlertjd:
Found a number of minor issues running integration tests with coreplus and S3, mostly around assuming that every storage provider is PyFsStorage.

### Review Checklist


tests/integration_tests/python/test_migrate_storage.py
```diff
@@ -9,20 +10,21 @@
 from api import config, util
 from bson.objectid import ObjectId
 
+def move_file(src_id, dst_storage, dst_path):
```
roycollins: If dst_storage is anything but a local fs it will break.  So it we later add a test migrating storage to S3 it will not work as the get_fs() method only exists on PyFs storage types. Not a problem right now just a point of caution as the abstraction layer grows.

ehlertjd: It's only used locally for setting up tests, and I think we could fairly easily set up the test in the other direction should the need arise.

PR #1553, opened 2019-03-04
[FLYW-485] Remove project and job-phi from the database

hkethi002:
### Changes
cleanup_deleted.py takes in new arguments to purge phi of a single project
- giving a project id (`--project`) flag will limit the cleanup to a single project, and will ignore the 72 hour requirement object to be deleted for before removing them from storage
- the `--job-phi` only works if the project_id is given. It will remove the `produced_metadata` field from each job whose destination is under that project. It will also remove all logs for the job, __and__ remove the container itself.

### Review Checklist


bin/cleanup_deleted.py
```diff
@@ -131,12 +204,28 @@ def cleanup_files(remove_all, origins):
...
-                    updated_doc = db.get_collection(container).update({'_id': document['_id']},
-                                                                      {'$pull': {'files': {'_id': f['_id']}}})
-                    if not updated_doc['nModified']:
-                        log.error('    couldn\'t remove file from database')
-                        exit(1)
+                    if not document_deleted:
+                        # only need to remove the file from the database
+                        # if the document wasn't already removed
+                        log.debug('    removing from database')
+                        updated_doc = db.get_collection(container).update({'_id': document['_id']},
```
ehlertjd: Use `update_one()` and `UpdateResult` instead?

bin/cleanup_deleted.py
```diff
@@ -56,24 +57,62 @@ def main(*argv):
...
+        Args:
+            container_list (list): list of ids of the destination containers
+
+        Returns:
+            tuple: tuple of modified count and deleted count
+    """
+    jobs = db.jobs.find({'destination.id': {'$in': [str(container_id) for container_id in container_list]}}, {'_id': 1})
+    log.debug('Found %s jobs', jobs.count())
+    job_requests = [pymongo.operations.UpdateOne({'_id': job['_id']}, {'$unset': {'produced_metadata': ''}}) for job in jobs]
+    jobs.rewind()
```
nagem: Neat.

bin/cleanup_deleted.py
```diff
@@ -56,24 +57,62 @@ def main(*argv):
...
+    job_log_requests = [pymongo.operations.DeleteOne({'_id': str(job['_id'])}) for job in jobs]
```
nagem: We'd want a delete_many here. job -> jobs_log is a one to many relationship.

bin/cleanup_deleted.py
```diff
@@ -111,9 +150,43 @@ def cleanup_files(remove_all, origins):
...
+            # Use the id field or parents.project field to filter results
+            # instead of date of deletion
+            project_filter = {
+                '$or': [
+                    {'_id': bson.ObjectId(project_id)},
+                    {'parents.project': bson.ObjectId(project_id)}
+                ]
+            }
+
+            # We don't care about time of deletetion for single project snipes
```
nagem: Make that point very clear in the args definition/help text

bin/cleanup_deleted.py
davidfarkas: As I see in the code there is only one `job_logs` document per `job`, and I suppose that the `_id` is unique so it will always match only one document, but correct me if I'm wrong. Here is the code snippet which add the logs:

```python
# Silently ignore adding no logs
if len(doc) <= 0:
   return

log = config.db.job_logs.find_one({'_id': _id})

if log is None: # Race
   config.db.job_logs.insert_one({'_id': _id, 'logs': []})

config.db.job_logs.update({'_id': _id}, {'$push':{'logs':{'$each':doc}}})
```

bin/cleanup_deleted.py
```diff
@@ -56,24 +57,62 @@ def main(*argv):
...
+        job_bulk_response = db.jobs.bulk_write(job_requests)
+        modified_count = job_bulk_response.modified_count
+    else:
+        modified_count = 0
+    if job_log_requests:
+        log.debug('going to remove %s job logs', len(job_log_requests))
+        job_log_bulk_response = db.job_logs.bulk_write(job_log_requests)
+        deleted_count = job_log_bulk_response.deleted_count
+    else:
+        deleted_count = 0
```
davidfarkas: Instead of bulk write I would prefer something like this:

```python
job_bulk_response = db.jobs.update_many({'_id': {'$in': jobs}}, {'$unset': {'produced_metadata': ''}})
job_log_bulk_response = db.job_logs.delete_many({'_id': {'$in': jobs}})
return job_bulk_response.modified_count, job_log_bulk_response.deleted_count
```

Or just using one `bulk_write` with these two methods.
With this you don't have to rewind and re-read the db again.
Is there any specific reason why you have used `bulk_write`?

bin/cleanup_deleted.py
hkethi002: I think you're right David, I'll revert those changes

bin/cleanup_deleted.py
hkethi002: The reason for the bulk write was so that mongo could chunk up the write operations in an optimal way. I can see the number of jobs for a project becoming very large eventually so this is an attempt to future proof the script in terms of scale.

I removed the the rewind call, thanks for catching that.

bin/cleanup_deleted.py
```diff
@@ -131,12 +233,32 @@ def cleanup_files(remove_all, origins):
...
+                job_log_operations += job_log_operations_increment
+                container_ids = []
+
+    if container_ids:
+        # find the jobs % 100 left
+        job_operations_increment, job_log_operations_increment = generate_job_operations(container_ids)
+        job_operations += job_operations_increment
+        job_log_operations += job_log_operations_increment
+        container_ids = []
+    jobs_modified, job_logs_deleted = execute_job_operations(job_operations, job_log_operations)
```
ehlertjd: We create them in increments of 100 but then execute them all at once?

PR #1554, opened 2019-03-04
Expose cli/release on version endpoint [FLYW-430]

mjburling:
### Review Checklist

When environment variables are missing, the values safely default to empty strings and appear thusly:

`{"applied_fixes": {"fix_subject_age_62": "2019-02-06T18:33:12.334000+00:00"}, "database": 63, "cli_version": "", "flywheel_release": "", "release": "", "_id": "version"}`

When the environment variables are provided, they render as expected:

`{"applied_fixes": {"fix_subject_age_62": "2019-02-06T18:33:12.334000+00:00"}, "database": 63, "cli_version": "7.1.2", "flywheel_release": "emerald.ci-dev.1", "release": "", "_id": "version"}`

Evaluated absence and presence on [test3](https://test3.dev.flywheel.io/api/version).

These changes depend on new logic in the installer to be introduced as part of emerald.7.0.0.


swagger/paths/version.yaml
```diff
@@ -7,8 +7,8 @@
         description: ''
         schema:
           $ref: schemas/output/version.json
-        examples: 
+        examples:
           response:
             _id: "version"
             database: 2
-          
+
```
hkethi002: 👌

swagger/schemas/definitions/version.json
```diff
@@ -7,11 +7,29 @@
         "_id": {
           "type": "string"
         },
+        "applied_fixes": {
+          "type": "object",
+          "description": "Collection of applied database fixes",
+          "patternProperties": {
+            "fix_([a-z0-9]+_)+[0-9]+$": { "type": "string" }
```
hkethi002: I don't know if this will fail CI if someone adds a malformed fix name, and then we'd have things in the db that wouldn't match the schema here

ehlertjd: You could just use `"additionalProperties": {"type": "string"}` instead, since swagger doesn't support `patternProperties`

mjburling: Is that a universal, or just this version of swagger? I want the 25 minutes of my life back.

ehlertjd: I think those 25 minutes [are gone](https://github.com/OAI/OpenAPI-Specification/issues/687#issuecomment-287806750)

swagger/paths/version.yaml
mjburling: I yield. EOF new lines are part of my vocabulary now.

swagger/schemas/definitions/version.json
mjburling: Implemented as suggested.

PR #1555, opened 2019-03-05
[FLYW-475] Regenerate and disable device API keys

jody-zeitler:
To satisfy the UI requirements of the devices page, API keys can be regenerated or disabled for a device (without deleting the device itself). This retains the behavior of on-demand key generation for bootstrapped devices such as the mongoconnector. Device are disabled by adding a `disabled` boolean to the record.

Swagger docs: https://flywheel-io.github.io/core/branches/device-keys/swagger/index.html

### Review Checklist


jody-zeitler:
Would it make more sense to disable a device rather than disable its key? I suppose I should also consider requests from disabled devices/keys to be invalid as part of the work as well.

ehlertjd:
@jody-zeitler
 
> Would it make more sense to disable a device rather than disable its key?
I think we can either disable or delete the key, but leave the device alone. AFAIK the use case here is that I can give out credentials temporarily for an import operation then revoke them, so it might make it easier for implementation to just delete them outright rather than adding a disabled flag.

 > I suppose I should also consider requests from disabled devices/keys to be invalid as part of the work as well.
Yeah, if the API key is disabled, we need to enforce that.

jody-zeitler:
@ehlertjd I can't delete a key outright, because it'll be regenerated automatically when the device is returned on get/list. I suppose I could distinguish between bootstrapped devices that require this behavior and those that do not, but I don't understand all the use cases of this behavior.

nagem:
@jody-zeitler @ehlertjd how does this proposal sound:

Site admins can "disable" devices. What that means in practice:
  - All existing API keys for that device are removed
  - Set a boolean of "disabled" on the device for filtering and display
  - Devices that are disabled cannot have an API key generated for them

To "Enable" a device:
  - Create a new API key for the device
  - Remove the disabled boolean (or set it to false)

ehlertjd:
@nagem Sounds good, I assume that the steps in "Enable" are transposed (i.e. enable *then* generate an API key)

jody-zeitler:
Works for me

api/handlers/devicehandler.py
```diff
@@ -62,7 +62,8 @@ def get_all(self):
     @staticmethod
     def join_api_key(device):
         api_key = DeviceApiKey.get(device['_id'])
-        device['key'] = api_key['_id'] if api_key else DeviceApiKey.generate(device['_id'])
+        if device.get('disabled', False) is False:
+            device['key'] = api_key['_id'] if api_key else DeviceApiKey.generate(device['_id'])
```
jody-zeitler: Should the key be `null` or omitted if the device is disabled?

ehlertjd: I think omitted should be fine based on the current schema (key is not a required field on device-output)

jody-zeitler:
@ehlertjd this is ready for another look

api/handlers/devicehandler.py
```diff
@@ -79,6 +80,27 @@ def post(self):
...
         return {'_id': result.inserted_id, 'key': key}
 
+    @require_admin
+    def put(self, device_id):
+        payload = self.request.json_body if self.request.body else {}
+
+        validate_data(payload, 'device-admin-update.json', 'input', 'PUT')
+        device = self.storage.get_container(device_id)
+        # Accept only a subset of device properties for update
+        update_payload = {prop: payload[prop] for prop in {'disabled'} if prop in payload}
```
ehlertjd: Couldn't we leverage the schema validation for this? (i.e. set `additionalProperties` to false and just take the update document?)

api/handlers/devicehandler.py
```diff
@@ -79,6 +80,27 @@ def post(self):
...
+        if len(update_payload) > 0:
```
ehlertjd: Use truthy checks in python `if update_payload:`

api/handlers/devicehandler.py
```diff
@@ -79,6 +80,27 @@ def post(self):
...
+            result = self.storage.update_el(device_id, update_payload)
+
+            is_disabled = update_payload.get('disabled', False)
+            if is_disabled is True:
```
ehlertjd: `if is_disabled:`
And similarly below `elif not is_disabled and device.get('disabled', False)`

swagger/schemas/definitions/device.json
```diff
@@ -48,6 +49,14 @@
       "x-sdk-model": "device",
       "additionalProperties": false
     },
+    "device-admin-update": {
+      "type": "object",
+      "properties": {
+        "disabled": {"type":"boolean"}
+      },
+      "x-sdk-model": "device",
+      "additionalProperties": true
```
ehlertjd: I would restrict additionalProperties here unless there's a good reason.

api/handlers/devicehandler.py
```diff
@@ -79,6 +80,27 @@ def post(self):
...
+                DeviceApiKey.revoke(device['_id'])
+            elif is_disabled is False and device.get('disabled', False) is True:
+                DeviceApiKey.generate(device['_id'])
+
+            return {'modified': result.modified_count}
```
ehlertjd: This is something that needs to be refactored globally, but I think we can skip the return here and just allow the http status to indicate success or failure.

swagger/schemas/definitions/device.json
```diff
@@ -34,7 +34,8 @@
         "errors":    {"$ref":"#/definitions/errors"},
         "info":      {"$ref":"common.json#/definitions/info"},
         "interval":  {"$ref":"#/definitions/interval"},
-        "last_seen": {"$ref":"common.json#/definitions/timestamp"}
+        "last_seen": {"$ref":"common.json#/definitions/timestamp"},
+        "disabled":  {"type":"boolean"}
```
ryandavisflywheel: I've seen a lot of times it makes sense to name things positively as a rule of thumb i.e. this should be "enabled" instead of "disabled".  This avoids a lot of weird logical gymnastics like where Justin was pointing out code above in his review... 
Instead of:
`if not is_disabled:`
we would have 
`if enabled:`
which is way easier to read and understand

Not being too familiar with the code base, I'm not gonna say change it because it may introduce issues elsewhere, but it may be a consideration.

ehlertjd: @ryandavisflywheel This is a good point, however for better or worse, we're following a [precedent with users](https://github.com/flywheel-io/core/blob/master/swagger/schemas/definitions/user.json#L34), so let's leave it as is for now.

ryandavisflywheel: Ah nice... I like consistency too :)

api/handlers/devicehandler.py
jody-zeitler: I was trying to decide whether to be restful or not. This endpoint should accept a full device record, but should document that everything but `disabled` is immutable.

api/handlers/devicehandler.py
jody-zeitler: I quote the Zen of Python https://www.python.org/dev/peps/pep-0020/

> Explicit is better than implicit.

jody-zeitler: plus it doesn't jive with my twisted javascript brain

roycollins: 👍 😛

ehlertjd: I will counter with [PEP-8](https://pep8.org/#programming-recommendations):
> Don’t compare boolean values to True or False using ==:
> Yes: `if greeting:`
> No: `if greeting == True:`
> Worse: `if greeting is True:`

jody-zeitler: Oh I fixed those. I had thought I was writing English.

api/handlers/devicehandler.py
```diff
@@ -86,20 +86,14 @@ def put(self, device_id):
...
-                DeviceApiKey.generate(device['_id'])
-
-            return {'modified': result.modified_count}
-        else:
-            return {'modified': 0}
+
+        result = self.storage.update_el(device_id, payload)
+
+        is_disabled = payload['disabled']
+        if is_disabled:
```
ehlertjd: 🎉

api/handlers/devicehandler.py
ehlertjd: That train of thought makes sense to me. We're effectively implementing `PATCH` on our `PUT` endpoints, so this is more a concern of consistency rather than being truly RESTful.

PR #1560, opened 2019-03-07
Log jobs and job logs [FLYW-507]

hkethi002:
#### Accessing a job directly logs:
- a `view_job` log with the job id and origin (of the access not the job)
- a `view_file` log for each file input in the `config.inputs` because the info object is in there
- a `view_container/subject` for each container found in the `produced_metadata` field
#### Accessing a job's logs logs:
- a `view_job_logs` log with the job_id and origin (of the access not the job)
#### Accessing a list of jobs (`/jobs` or `/sessions/*/jobs`)
Doesn't log anything but we strip out the info object for inputs and the produced metadata
### Review Checklist


api/jobs/job_util.py
```diff
@@ -5,9 +5,23 @@
 from ..dao.basecontainerstorage import ContainerStorage
 from ..dao.containerutil import singularize
 
+
+def clean_job(job_map):
+    """Remove certain fields from jobs to simplify the endpoint, the fiels
```
mjburling: Just a typo at the end of the line.

mjburling:
Walked through the functionality with Harsha. He's missing at least some tests at this point, but this appears to effectively reduce the risk of unnecessary PHI exposure while providing log access.

api/jobs/job_util.py
```diff
@@ -5,9 +5,23 @@
...
+    are produced metadata and info objects on config.inputs items
+
+    Args:
+        job_map  (dict): A job object to be returned
+    """
```
kofalt: Please change this function to return a cleaned copy of the job, rather than destructively modify.

api/jobs/job_util.py
```diff
@@ -5,9 +5,23 @@
...
+    job_map.pop('produced_metadata', None)
+    for config_input in job_map['config'].get('inputs', {}).values():
+        if config_input.get('base') == 'file':
+            config_input['object'].pop('info', None)
```
nagem: Can we just not return these keys via a projection?

hkethi002: The key is technically arbitrary, i.e. it might be `config.inputs.dicom.object` if `config.inputs.dicom.base` is `file`, where `dicom` could be arbitrary.

PR #1564, opened 2019-03-14
Set providers on jobs [FLYW-433]

ehlertjd:
In short, this starts making use of the site settings and providers functionality, behind a `multiproject` feature flag.

## Fixed Mode
This is the default behavior, where all jobs are unconditionally assigned to the static site provider. This therefore is a non-breaking change from a functionality perspective.

Compute provider can still be overridden in this mode.

## Multiproject Mode
To enable multiproject mode, the following environment variable must be set: `SCITRAN_PERSISTENT_MULTIPROJECT=true`. In this mode, whenever a job is scheduled, a compute provider will be selected according to the table below, with the following notable exceptions:
- If a job is retried, it will use the original jobs provider
- If an override provider is specified (see endpoint documentation) that provider will be used

Provider selection rules:

Gear Type | Some inputs are device inputs | No inputs are device inputs
---------- | ------------------------------ | ---------------------------
Gear is "Center Pays" gear | Center's Compute Provider | Project's Compute Provider
Gear is not "Center Pays" gear | Project's Compute Provider | Project's Compute Provider

### Review Checklist


tests/unit_tests/python/test_provider_pickers.py
```diff
@@ -0,0 +1,209 @@
...
+    assert provider_id == group_provider_id
+
+    # == Project ==
+    MockContainerStorage.get_el.side_effect = (
+        {'_id': 'project', 'providers': {'compute': project_provider_id}},
+        {'_id': 'group', 'providers': {'compute': group_provider_id}}
+    )
+    is_site, provider_id = picker.get_provider_id_for_container(container, cls)
+    assert is_site == False
+    assert provider_id == project_provider_id
```
hkethi002: I don't think we _need_ to test the case where project provider is set but group provider isn't.

api/jobs/jobs.py
```diff
@@ -64,6 +64,10 @@ def __init__(self, gear, inputs, destination=None, tags=None,
         related_container_ids: list (optional)
             The set of all container ids related to inputs and destination of this job, as of
             job creation time. This field is not updated when containers are moved.
+        label: str (optional)
+            An optional label for the job
+        compute_provider_id: ObjectId (optional)
```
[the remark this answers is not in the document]
ehlertjd: Good point, we're still multiprocess so I think we still need the safe ID generation. Historically there have been attempts to start transitioning to uuids on a case-by-case basis, but that resulted in numerous defects. For the moment I'd like to hold the line on ObjectIds but it is desirable to move to opaque strings when we can convert everything at once.

tests/unit_tests/python/test_provider_pickers.py
ehlertjd: I agree. This is testing that we return the project provider when _both_ group and project are set.

ehlertjd:
@hkethi002 I've added test cases in aeb9e04 to cover your suggestions on batch testing!

PR #1566, opened 2019-03-14
Formalize feature flags from environment variables [FLYW-513]

ehlertjd:
Translates environment variables in the form `FLYWHEEL_FEATURE_FEATURE_NAME=value` to `"feature_name": "value"` on the `/api/config` endpoint. Case-insensitive values `true` or `false` will be converted to the corresponding boolean value.

### Review Checklist


roycollins:
Did we want to consider moving the feature flags to the persistence layer?  Then we could turn feature flags on and off directly without needing operations to access the servers.  If we put them in the singletons collection on Mongo we could read that and see if the flag is set each time the code executes.  I am not sure if gunicorn will see env var changes while running but I am guessing it will require are a restart of the service.  Just a thought to get us closer to zero downtime configuration changes.  Otherwise it looks good to me.

ehlertjd:
@roycollins You can set features in the database. I added [one more test clause](https://github.com/flywheel-io/core/pull/1566/commits/5302c568d13cbb8c365549d1262c771a3f8c935e) to demonstrate that.

PR #1572, opened 2019-03-20
More tolerant config, log job on analyses [FLYW-613]

hkethi002:
### Review Checklist


api/web/base.py
```diff
@@ -52,6 +52,36 @@ def initialize(self, request, response):
...
+        """Log a view_file access for each file input for the job because we
+        are going to return the info object on the inputs
+
+        Args:
+            job (Job): A job object
+        """
+        if job.config:
+            for config_input in job.config.get('inputs', {}).values():
+                if config_input.get('base') == 'file':
+                    file_parent_type = config_input['hierarchy']['type']
```
ehlertjd: Since we're questioning everything - do we know that all `file` inputs have fully populated `hierarchy` and `location` subdocs?

hkethi002: @nagem and I had a conversation about this, and we decided to draw the line at the config inputs are probably safe, but maybe we should surround it in try except statements and just skip over it.

PR #1575, opened 2019-03-25
Handle edge cases in job detail [FLYW-649]

ehlertjd:
### Review Checklist


api/jobs/handlers.py
```diff
@@ -624,6 +624,8 @@ def get_detail(self, _id):
         def get_container(ref):
             """Helper for cached retrieval of container"""
             # Normalize id
+            if not ref or not ref.id:
+                return None
```
hkethi002: Out of curiosity, how does this happen?

nagem:
¯\\_(ツ)_/👍

api/jobs/jobs.py
```diff
@@ -159,7 +159,7 @@ def load(cls, e):
         # Don't modify the map
         d = copy.deepcopy(e)
 
-        if d.get('inputs'):
+        if d.get('inputs') is not None:
```
nagem: I'm not sure I understand this one - you want to still run through this code path if inputs is `{}`?

nagem: Or were inputs empty lists sometimes?

hkethi002: It's `[]` in mongo and we still want to convert it into a dictionary

api/jobs/handlers.py
ehlertjd: Thank you for asking! This was happening when the job destination was not set or None. However when I added the missing integration test I uncovered another NPE if inputs exist, but destination is `None` which is not something we observed in the database, but can't rule it out.

ehlertjd:
@roycollins Good thought, but that already exists in the current codebase. This fix is just to handle ancient jobs which didn't have those restrictions.

PR #1576, opened 2019-03-26
Storage providers

roycollins:
Creates storage providers using the provider abstraction layer.  Each file is now assume to belong to a single provider.  There is an upgrade script that will have to run to set the provider_id on each file in the database to be the new default site provider.   

Please review the test carefully as the assumptions changed so the tests cases were also adjusted.  There is no longer a legacy type file system for the CAS structure.  It is now assumed that all files will have an id, along with the new provider_id.   

I am still going to review the test cases and add more coverage but we can start the review process. 
### Review Checklist


api/config.py
```diff
@@ -338,10 +348,12 @@ def get_release_version():
...
+#primary_storage = create_flywheel_fs(__config['persistent']['fs_url'])
+# primary_storage = create_flywheel_fs(__config['persistent']['data_path'])
+
+# This is used for the temp storage and must be set
+local_fs_url = __config['persistent']['data_path']
 
-storage = create_flywheel_fs(__config['persistent']['fs_url'])
-# local_fs must be PyFS with osfs for using the local get_fs functions for file manipulation
-local_fs = create_flywheel_fs('osfs://' + __config['persistent']['data_path'])
 support_legacy_fs = __config['persistent']['support_legacy_fs']
```
ehlertjd: I think 357-372 should be removed now.

api/data_export/strategy/full.py
```diff
@@ -136,5 +136,5 @@ def _create_metadata_target(self, path, parent_type, parent_id, container, file_
 
         # NOTE: We always return size=0 for metadata - it will be set just-in-time
         return models.DownloadTarget('metadata_sidecar', dst_path, parent_type, parent_id,
-            container['modified'], 0, 'metadata', file_id=file_id, filename=filename,
+            container['modified'], 0, 'metadata', provider_id, file_id=file_id, filename=filename,
```
ehlertjd: You can use a constant `None` here, the metadata will always be pulled from the DB instead of storage.

api/config.py
```diff
@@ -122,7 +121,18 @@ def apply_env_variables(config):
...
+    # TODO: These shold be static constants from the provider class but this creates ciruclar 
+    # dependencies on the import ordering
+    signed_url = False
+    if db.providers.find_one({
+        'provider_class': 'storage',
+        'provider_type': {'$in' : ['aws','gc']}
+        }):
+        signed_url = True
+
+    config['features']['signed_url'] = signed_url
```
ehlertjd: This should be `True` iff **all** providers support signed urls, (with the exception of temp storage)

api/site/providers/aws_storage_provider.py
```diff
@@ -0,0 +1,242 @@
...
+        return self._storage_url
+
+    @property
+    def storage_plugin(self):
+        """
+            Allow access to the internal storage provider
+        """
+        return self._storage_plugin
+
+    def _make_request(self, method, request_parameters, host):
```
ehlertjd: Is this something we could delegate to the [storage Interface](https://github.com/flywheel-io/flywheel-common/blob/master/flywheel_common/storage/interface.py#L3) for storage providers? I think it would make sense to leverage the [boto client](https://boto3.amazonaws.com/v1/documentation/api/latest/index.html) to generate requests.

api/config.py
roycollins: Yes, fixed.

api/site/providers/aws_storage_provider.py
roycollins: I agree and I would have loved to use the boto client to speed up development. The reason I did not use the client was to be consistent with GC Compute provider where you did not want to include the client SDK for just a couple of API calls to verify permissions.  I assumed the same was true here. It a minor change either way but I just need clarity on when the project should use the vendor libs and when they should not.

api/data_export/strategy/full.py
roycollins: fixed

bin/oneoffs/migrate_storage.py
```diff
@@ -206,77 +188,42 @@ def f_retry(*args, **kwargs):
...
-            # We dont support delete at this time so we should check the hash before we save the file if we dont extra files lying around
-            # target_fs.remove(f_new_path)
-
-
-def migrate_analysis_file(f, migrated_files):
-    match = [cf for cf in migrated_files if
-             cf['fileinfo']['hash'] == f['fileinfo']['hash'] and cf['fileinfo'].get('_id')]
-    # The file is already migrated
-    if len(match) > 0 and not f['fileinfo'].get('_id'):
-        log.debug('    file was already migrated, just set a reference')
```
hkethi002: Why do we not need to check for this anymore?

bin/database.py
```diff
@@ -2344,6 +2344,33 @@ def upgrade_to_63():
...
+def upgrade_to_64():
+    '''
+    All project templates should be list of templates
+    '''
+
+    config.db.create_collection('providers')
+
+    provider = config.db.providers.insert_one({
+        "origin": {"type":"system","id":"system"},
+        "created":"2019-03-19T18:48:37.790Z",
```
hkethi002: Is this supposed to be hardcoded, or `datetime.datetime.now()`

bin/database.py
```diff
@@ -2344,6 +2344,33 @@ def upgrade_to_63():
 ### END RESERVED UPGRADE SECTION
 ###
 
+def upgrade_to_64():
+    '''
+    All project templates should be list of templates
```
hkethi002: I think this is supposed to be changed

tests/unit_tests/python/test_config.py
```diff
@@ -113,3 +114,39 @@ def test_create_or_recreate_ttl_index(mocker):
     db[collection].drop_index.assert_called_with(index_id)
     db[collection].create_index.assert_called_with(index_name, expireAfterSeconds=ttl)
 
+def test_signed_urls():
+    regex = re.compile("remove_me")
+
+    #We should assume our tests start with a valid state
+    api.config.db.providers.remove({'label': regex})
+    api.config.__last_update = datetime.datetime(2000, 01, 01)
```
ehlertjd: This is red because it'll be invalid syntax when we move to python3.

roycollins:
Superseded by #1597

PR #1583, opened 2019-04-08
Save search permissions [FLYW-688]

hkethi002:
Functionality-wise, permissions on saved search queries follow data views. 

### Review Checklist


api/dao/containerstorage.py
```diff
@@ -528,11 +528,90 @@ def inflate_job_info(self, analysis, remove_phi=False):
...
+        """
+        # Everyone has access to read queries on the site and on themselves
+        parents = ['site', uid]
+
+        # Get the projects they have access (ids as strings)
+        parents += [str(p['_id']) for p in ProjectStorage().get_all_el(None, {'_id': uid}, {'_id': 1})]
+
+        # Get the groups that user has access to
+        parents += [g['_id'] for g in GroupStorage().get_all_el(None, {'_id': uid}, {'_id': 1})]
+
```
nagem: What amount of overlap is there between this and the DataViews code? Might make sense to create a shared abstraction above the two at some point.

api/dao/containerstorage.py
```diff
@@ -528,11 +528,90 @@ def inflate_job_info(self, analysis, remove_phi=False):
...
+                'parent.id': {
+                    '$in': [
+                        parent_id for
+                        parent_id in
+                        self._get_user_parents(user['_id'])
+                    ]
+                }
+            })
+            log.debug('The query is')
+            log.debug(query)
```
nagem: Some debug statements came along for the ride.

api/handlers/dataexplorerhandler.py
```diff
@@ -994,38 +994,78 @@ def __init__(self, request=None, response=None):
...
+        parent_container = self.storage.get_parent(None, cont=payload)
+        self.permcheck('POST', parent_container=parent_container)
+
         result = self.storage.create_el(payload)
         if result.inserted_id:
             return {'_id': result.inserted_id}
         else:
             raise APIStorageException("Failed to save the search")
 
     def get_all(self):
```
nagem: Can we add "require_login" to all of these? @ehlertjd how do you feel about handling non-authenticated requests in our current state of not _really_ supporting them?

api/dao/containerstorage.py
hkethi002: There isn't a whole lot of overlap, the main overlap I can think of is the [permission checking](https://github.com/flywheel-io/core/pull/1583/files#diff-9a96bfba10be2c47cfd6d78dc7e84d0cR1057). The main difference happens with how the parents are handled between the two, the save searches more closely resembles analyses, so I guess it would be worth seeing if it makes sense to inherit the referer handler.

tests/integration_tests/python/test_saving_search_queries.py
```diff
@@ -33,38 +46,65 @@ def test_saving_search_queries(as_admin, data_builder):
...
 
     # Delete saved search
     r = as_admin.delete('/dataexplorer/queries/' + search)
     assert r.ok
     r = as_admin.get('/dataexplorer/queries')
     assert r.ok
     assert len(r.json()) == 0
+
+
+def test_saving_site_search_queries(as_admin, data_builder, as_user):
```
ehlertjd: What classes of users can view/ save / delete site queries? Can we test those assumptions here?
i.e.
- public view/edit/delete
- users view/edit/delete
- admins view/edit/delete

api/dao/containerstorage.py
```diff
@@ -528,11 +528,87 @@ def inflate_job_info(self, analysis, remove_phi=False):
...
+            uid (str): The id of the user to get parent ids for
+            user_is_admin (bool): User is or isn't site admin
+
+        Returns:
+            list: list of ids of parents as strings
+        """
+        # Everyone has access to read queries on the site and on themselves
+        parents = [uid]
+        if user_is_admin:
+            parents.append('site')
```
ehlertjd: Seems like users should be able to see and execute but not edit site queries, no?

hkethi002: The container referer auth requires admin for any interaction with site, so I thought it would make sense to have the get all respect that as well

PR #1586, opened 2019-04-10
Add drone login to python SDK [FLYW-682]

ehlertjd:
Intended use is for internal flywheel services that still make use of the drone key (e.g. condor).

### Review Checklist


hkethi002:
Would it make sense for the drone login to automatically retrieve its api key and start using that instead.

ehlertjd:
@hkethi002 yes, that's probably the "correct" way to do this.

PR #1588, opened 2019-04-11
Provider on job preview [FLYW-694]

ehlertjd:
Determines what the effective provider for a job should be, and returns that provider's record.
When using for analysis gears, set the destination as the parent container of the analysis.

### Review Checklist


api/jobs/handlers.py
```diff
@@ -560,6 +561,23 @@ def add(self):
 
         return { '_id': job.id_ }
 
+    @require_login
+    def determine_provider(self):
+        """Determine the effective provider for a job"""
+        payload = self.request.json
```
hkethi002: Should we be validating the job input here against the schema?

ehlertjd: Less worried about it here than in the [add endpoint](https://github.com/flywheel-io/core/blob/78d3ab00976c0e8482586c8ec9aef61caf86325e/api/jobs/handlers.py#L547-L562). At least here we don't put anything into the database.

hkethi002: Ah, I forgot that enqueueing a job doesn't actually enqueue it

ehlertjd:
Good point, lets address that in a separate refactor PR

PR #1589, opened 2019-04-12
Allow compute-provider on job ask [FLYW-714]

ehlertjd:
### Review Checklist


api/jobs/queue.py
```diff
@@ -386,11 +387,13 @@ def ask(query):
                 "group": [],
                 "gear-name": [],
                 "tag": [],
+                "compute-provider": [],
```
roycollins: Perhaps call it compute_providers since its a list and it seems we can have more than one.

api/config.py
```diff
@@ -223,6 +223,7 @@ def initialize_db():
     db.jobs.create_index('tags', **kwargs)
     db.jobs.create_index([('destination.type', 1), ('destination.id', 1)], **kwargs)
     db.jobs.create_index([('inputs.type', 1), ('inputs.id', 1)], **kwargs)
+    db.jobs.create_index('compute_provider_id', **kwargs)
```
kofalt: This will make our tenth index. Many of these indexes can be assured of existing for every job.

Our largest customer, IIRC, has something on the order of a quarter-million job records. ~Houston, what's the word on these panels shorting out?~ Justin, what's the word on this causing mongo or disk pressure at high running job counts?

ehlertjd: I don't think we're in the danger zone just yet. On one site with 270k jobs:
```
	"totalIndexSize" : 26542080,
	"indexSizes" : {
		"_id_" : 3100672,
		"inputs.id_1_inputs.type_1" : 1708032,
		"state_1_now_1_modified_1" : 2535424,
		"related_container_ids_1" : 6066176,
		"tags_1" : 3231744,
		"parents_1" : 1748992,
		"destination.type_1_destination.id_1" : 1597440,
		"inputs.type_1_inputs.id_1" : 1568768,
		"created_1" : 2486272,
		"modified_1" : 2498560
	}
```
So roughly 26 MB of indexes, and adding another id for each field, so I'd estimate that it brings us to ~30MB.

kofalt: I was thinking more about the number of operations the DB has to perform when an update happens, but since that's so tiny on disk I'm going to assume it's trivial.

PR #1591, opened 2019-04-15
Update PR template

ehlertjd:
Added a note about DB upgrades and test plans.

### Review Checklist


nagem:
I like the list, let's hold each other accountable to it!

PR #1599, opened 2019-04-26
Add file modality to list of aggregations to support faceted search [FLYW-788]

nagem:
Add `file.modality` in addition to `file.type` and `file.classification`. 

### Review Checklist


nagem:
Please let me know if we have a proper way to test the facet aggregations.

PR #1601, opened 2019-04-26
Improved job stats

ehlertjd:
Use estimated document count instead of aggregation for jobs/stats

### Review Checklist

### DIFF
--- a/api/jobs/queue.py
+++ b/api/jobs/queue.py
@@ -469,18 +469,12 @@ class Queue(object):
 
         query = Queue.lists_to_query(whitelist, blacklist, capabilities)
 
-        # Pipeline aggregation
-        result = list(config.db.jobs.aggregate([
-            {'$match': query },
-            {'$group': {
-                '_id': '$state',
-                'count': {'$sum': 1}}
-            }
-        ]))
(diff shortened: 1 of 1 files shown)


api/jobs/queue.py
```diff
@@ -469,18 +469,12 @@ def job_states(whitelist, blacklist, capabilities):
...
-
-        # Map the mongo result to something useful
-        by_state = {s: 0 for s in JOB_STATES}
-        by_state.update({r['_id']: r['count'] for r in result})
+        # Use count documents for faster resolution
+        by_state = {}
+        query_copy = copy.deepcopy(query)
+        for state in JOB_STATES:
+            query_copy['state'] = state
+            by_state[state] = config.db.jobs.estimated_document_count(query=query_copy)
```
hkethi002: [MongoDb](https://docs.mongodb.com/manual/reference/method/db.collection.estimatedDocumentCount/#behavior) is saying that it doesn't take a query filter even though we're passing one in here. Does this apply the filter?

ehlertjd: It does, this is live on dev right now.

PR #1602, opened 2019-05-01
Loaded rules are lists instead of None [FLYW-806]

hkethi002:
We iterate over certain fields of a rule, and if they are null in the db, we end up with an error, so we should load them in as a list if they are None.

### Review Checklist


api/jobs/mappers/rules_mapper.py
```diff
@@ -132,5 +132,9 @@ def _load_rule(self, rule_doc):
         if rule_doc is None:
             return None
 
+        rule_doc['any'] = rule_doc.get('any') or []
```
nagem: The "pythonic" way to write this would be `rule_doc.get('any', [])`

nagem: Or the entire line could be `rule_doc.setdefault('any', [])`

PR #1603, opened 2019-05-02
Use uid + mail_domain to synthesize mail

ehlertjd:
### Review Checklist


tests/unit_tests/python/test_auth.py
```diff
@@ -99,6 +99,36 @@ def test_jwt_auth_with_mail_format(config, as_drone, as_public, api_db):
...
+        check_ssl=False)
+
+    uid = 'ldapuser'
+    email = 'ldapuser@scitran.edu'
+    api_db.users.delete_one({'_id': email})
+    with requests_mock.Mocker() as m:
+        # try to log in w/ ldap
+        assert as_drone.post('/users', json={'_id': email, 'firstname': 'test', 'lastname': 'test'}).ok
+
+        m.post(config.auth.ldap.verify_endpoint, json={'uid': uid, 'mail': 'ldap@scitran.edu'})
```
nagem: @ehlertjd do we need to modify the JWT service to return both of these keys?

ehlertjd:
Went a different route

PR #1477, opened 2018-12-12
[FLYW-43] Center flag

hkethi002:
Ref #1183 

### Review Checklist

- Tests were added to cover all code changes
- Documentation was added / updated
- Code and tests follow standards in CONTRIBUTING.md

PR #1605, opened 2019-05-06
[FLYW-43] Group lab edition

roycollins:
Add lab edition settings for groups and projects.

Lets verify the tests are accurate with regard to the business logic around when is_lab can change and what values should be propagated throughout. 

One note is that this PR will need the modified providers branch to run successfully. 

### Review Checklist


api/handlers/containerhandler.py
```diff
@@ -325,6 +325,12 @@ def post(self, cont_name):
             # Validate any changes to storage providers
             providers.validate_provider_updates({}, payload.get('providers'), self.user_is_admin)
 
+            if 'is_lab' in payload and not self.user_is_admin:
+                self.abort(403, 'Only admins can specify lab_edition')
```
jody-zeitler: Only admins can specify is_lab

api/handlers/containerhandler.py
```diff
@@ -358,6 +364,8 @@ def put(self, cont_name, **kwargs):
                 r_payload[key] = payload[key]
 
         if cont_name == 'projects':
+            if 'is_lab' in payload and not self.user_is_admin:
+               self.abort(403, 'Only admins can specify lab_edition')
```
jody-zeitler: Only admins can specify is_lab

api/handlers/containerhandler.py
roycollins: Good call.   Fixed those!

ehlertjd:
I think we still want to make use of the `editions` field as discussed in prior iterations of this feature: #1183 - Lab group flag and #1477 - Center flag

I think those PRs are a good place to start on getting some background for this feature.

roycollins:
I have converted the format of the model to use the editions array.  Please let me know if this looks to be the desired approach

roycollins:
Superseded by https://github.com/flywheel-io/core/pull/1647

PR #1610, opened 2019-05-09
Add checks module and check for CAS [FLYW-846]

ehlertjd:
### Review Checklist


bin/checks.py
```diff
@@ -0,0 +1,75 @@
...
+    result = globals().get(check_id, None)
+    if not result:
+        raise ValueError('Unknown check method: {}'.format(check_id))
+    return result
+
+def check_for_cas_files():
+    """
+    Check that all CAS files have been migrated in a system.
+    """
+    for collection_name in [ 'acquisitions', 'sessions', 'subjects', 'projects', 'collections' ]:
```
nagem: To be safe you should also have analyses here.

PR #1619, opened 2019-05-15
Add worker process to collect metrics [FLYW-793]

ehlertjd:
No handling of gunicorn's graceful code reload, but that keeps the logic simple.

### Review Checklist


ehlertjd:
@kofalt For context, `livesum` is what we used back in the uwsgi mule worker days (i.e. single process)

PR #1620, opened 2019-05-16
Collect mongo statistics in metrics

ehlertjd:
This will collect dbstats as well as more detailed collection stats. May assist in determining when and how to scale elastic.

### Review Checklist


api/metrics/collect.py
```diff
@@ -32,15 +28,27 @@ def collect_db_metrics():
...
         for state in JOB_STATES:
             count = config.db.jobs.count({'state': state})
             values.JOBS_BY_STATE.labels(state).set(count)
 
-        # Get raw collection counts
-        for collection_name in COUNT_COLLECTIONS:
-            count = config.db[collection_name].count()
-            values.COLLECTION_COUNT.labels(collection_name).set(count)
+        # Get raw collection stats
+        for collection_name in config.db.list_collection_names():
```
roycollins: We might want to keep this to a subset of collections we care about.  The full list will have collections that dont change very often. Each collection will require a full call to mongodb every 30 seconds.  Perhaps just db.stats() will give enough info?

ehlertjd: The stats commands appear to use cached counts from mongodb, so I don't think expense is an issue. I'd rather have a few more collections captured than be missing one when we need it.

PR #1621, opened 2019-05-17
Add endpoint to check for existence of UIDs

ehlertjd:
### Review Checklist


ehlertjd:
We are [only searching what you ask for](https://github.com/flywheel-io/core/pull/1621/files#diff-4c6cc92be5eba46dedd0d077002ba72bR21). It's just not explicit 🤷‍♂

PR #1623, opened 2019-05-21
Fix false positive CAS check on empty file array

ehlertjd:
Use same style check (`$elemMatch`) as the [storage providers upgrade](https://github.com/flywheel-io/core/pull/1597/files#diff-837f544139dd99be7b888e157752643bR2448).

### Review Checklist


ehlertjd:
Good reference, thank you

PR #1625, opened 2019-05-21
Handle RFC2231 encoded filenames [FLYW-882]

ehlertjd:
Ref: https://github.com/kennethreitz/requests/issues/2117
Ref: https://github.com/kennethreitz/requests/issues/2313

Also Ref: https://github.com/urllib3/urllib3/blob/1.24.3/src/urllib3/fields.py#L45
And: [StackOverflow](https://stackoverflow.com/questions/18094309/decoding-rfc-2231-headers) for API usage.

### Review Checklist


tests/integration_tests/python/test_uploads.py
```diff
@@ -2007,15 +2007,8 @@ def test_unicode_filenames(data_builder, file_form, as_admin, as_public, api_db)
...
-    if core_works:
-        r = as_admin.post('/projects/' + project + '/files', files=file_form((filename, file_contents)))
-        assert r.ok
-    else:
-        r = as_admin.post('/projects/' + project + '/files', files=file_form(('placeholder.csv', file_contents)))
-        assert r.ok
-
-        api_db.projects.update_one({'_id': bson.ObjectId(project)}, {'$set': {'files.0.name': filename}})
+    r = as_admin.post('/projects/' + project + '/files', files=file_form((filename, file_contents)))
+    assert r.ok
```
hkethi002: If it's possible to send a blank filename, could we test that?

ehlertjd: Is this the appropriate PR to do this? And what's the expected result?

ehlertjd: Testing revealed that this does not alter behavior for blank filenames. I think we can raise a separate issue.

PR #1626, opened 2019-05-21
[FLYW-898] Add pagination option to collection/sessions query

jody-zeitler:
https://flywheelio.atlassian.net/browse/FLYW-898

I need this for holistic session pagination in the UI. I would use the root /sessions endpoint but joining sessions to collections is weird, so I opted for the less awkward approach.

### Review Checklist


ehlertjd:
Please add [x-fw-pagination: true](https://github.com/flywheel-io/core/blob/master/swagger/paths/collections.yaml#L11) to the [appropriate endpoint](https://github.com/flywheel-io/core/blob/master/swagger/paths/collections.yaml#L106) in the swagger docs. Otherwise LGTM

PR #1627, opened 2019-05-22
Update migrate_storage.py

roycollins:
### Review Checklist


roycollins:
It should be back ported to releases. It will prevent CAS migration indirectly if you are missing files. But even with this fix in place the migration stops on the missing file, at least it will give a clean error with this fix.  I am not sure if we should skip files that are missing at the minimum we need to update the DB so they are not seen as CAS files anymore.  Just put in a stub UUID?

ehlertjd:
I think missing files should be dealt with on a case-by-case basis; I don't think that's a problem we can (nor should try) to solve with software alone.

ehlertjd:
Closed for #1643

PR #1628, opened 2019-05-22
[FLYW-844] [FLYW-888] Session pagination filter considerations

jody-zeitler:
### [FLYW-844]

Filters were choking with numbers as regex filter values (e.g. `filters=label=~100`) because they were being coerced to numbers and the Mongo $regex operators only works on strings. This change forgoes value coercion on regex filters and keeps them as strings.

### [FLYW-888]

Adding a boolean value case for filter value coercion, so we can filter sessions with `satisfies_template!=false` (true or undefined) and `satisfies_template=false` to partition flagged sessions in the UI.

### Review Checklist

[FLYW-844]: https://flywheelio.atlassian.net/browse/FLYW-844
[FLYW-888]: https://flywheelio.atlassian.net/browse/FLYW-888


api/util.py
```diff
@@ -420,7 +420,7 @@ class PaginationParseError(ValueError):
 
 
 def parse_pagination_value(value):
-    """Return casted value (ObjectId|datetime|float) from user input (str) for use in mongo queries."""
+    """Return casted value (ObjectId|datetime|float|boolean) from user input (str) for use in mongo queries."""
     if len(value) > 1 and value[0] == '"' and value[-1] == '"':
         return value[1:-1]
     elif bson.ObjectId.is_valid(value):
```
ehlertjd: This reminded me that [12-byte strings are treated as ObjectIds](https://github.com/flywheel-io/core/pull/1529#issuecomment-461614984). We should probably test and remedy that.

PR #1631, opened 2019-06-03
User auth tokens

davidfarkas:
With this PR user's will be able to connect multiple google accounts to Flywheel. This feature is used in the Google Healthcare integration. We use user's google access token to retrieve data from Healthcare API. I used the already existing auth providers , so later it is possible to extend it to other third-party ones.

### Review Checklist


api/handlers/userhandler.py
```diff
@@ -244,3 +246,70 @@ def get_jobs(self):
...
+        token = config.db.user_auth_tokens.find_one({'_id': _id, 'uid': self.uid})
+        if token:
+            auth_provider = AuthProvider.factory(token['auth_type'])
+            auth_provider.revoke_token(token['access_token'])
+            # Revoke the refreshtoken too if it is specified
+            if token.get('refresh_token'):
+                auth_provider.refresh_token(token['refresh_token'])
+            # Delete token from our database
+            config.db.user_auth_tokens.delete_one({'_id': ObjectId(_id), 'uid': self.uid})
+            return {'deleted': 1}
```
hkethi002: I would check the response and make sure that the deleted count is in fact 1 and otherwise raise an exception

PR #1632, opened 2019-06-04
Add basic authentication provider

ehlertjd:
This provider is for development only.
NOT SUITABLE FOR PRODUCTION USE!!

Also introduces a metric for auth_provider types on site.

### Review Checklist


tests/unit_tests/python/test_auth.py
```diff
@@ -483,3 +483,93 @@ def test_saml_auth(config, as_drone, as_public, api_db):
...
+        # Attempt Set password
+        r = as_public.put('/users/' + uid, json={'password': 'changeme'})
+        assert r.status_code == 403
+
+        # Cannot reconfigure password if basic auth not configured
+        basic_cfg = config['auth'].pop('basic')
+        r = as_drone.put('/users/' + uid, json={'password': 'changeme'})
+        assert r.status_code == 400
+
+        # Change password (as admin)
```
hkethi002: We should test that the user can change their own password with admin

api/auth/authproviders.py
```diff
@@ -85,6 +88,73 @@ def set_refresh_token_if_exists(self, uid, refresh_token):
...
+        config.db.users.update_one({'_id': uid}, {'$set': {
+            'password_hash': self.password_hasher.hash(password),
+            'modified': datetime.datetime.now()
+        }})
+
+    def validate_code(self, code, **kwargs):
+        # Code is payload with email, password fields
+        uid = self.validate_user(code['email'], code['password'])
+        return {
+            'access_token': base64.b64encode(os.urandom(42)),
```
kofalt: It looks like there are places (like this) that use b64 directly instead of calling `util.create_nonce`. Would you agree that this helper func should be used instead of scattered `urandom` stuff?

api/handlers/userhandler.py
```diff
@@ -84,12 +85,24 @@ def put(self, _id):
         payload = self.request.json_body
         if not payload:
             self.abort(400, 'PUT request body cannot be empty')
+
+        # Password is only valid if basic auth enabled
+        password = payload.pop('password', None)
```
kofalt: Why not have dedicate password routes instead, which only work if the basic provider is enabled?

api/handlers/userhandler.py
```diff
@@ -84,12 +85,24 @@ def put(self, _id):
...
         payload_validator = validators.from_schema_path(payload_schema_uri)
         payload_validator(payload, 'PUT')
 
+        if password is not None:
+            try:
+                auth_provider = AuthProvider.factory('basic')
+            except NotImplementedError as e:
+                self.abort(400, str(e))
+
+            payload['password_hash'] = auth_provider.hash(password)
```
kofalt: Wouldn't this logic run even if the basic provider is not enabled? What am I missing?

api/auth/authproviders.py
```diff
@@ -85,6 +88,73 @@ def set_refresh_token_if_exists(self, uid, refresh_token):
...
+        self.password_hasher = argon2.PasswordHasher(**kwargs)
+
+    def hash(self, password):
+        """Return a hashed password"""
+        return self.password_hasher.hash(password)
+
+    def set_user_hash(self, uid, password):
+        """Update a user's hash in the database"""
+        config.db.users.update_one({'_id': uid}, {'$set': {
+            'password_hash': self.password_hasher.hash(password),
```
kofalt: Could we please do a `passwords` or `user_passwords` collection instead?

api/auth/authproviders.py
ehlertjd: I would agree that we should have a shared function for generating secure random byte strings. `util.create_nonce` is not that function.

ehlertjd:
@kofalt A lot of good comments to be addressed when/if we turn this into a production feature. For now I'd much prefer to limit the scope of this in order to get it merged and usable for automated QA validation of our sites.

api/auth/authproviders.py
```diff
@@ -85,6 +88,73 @@ def set_refresh_token_if_exists(self, uid, refresh_token):
...
+        super(BasicAuthProvider, self).__init__('basic')
+
+        # Create PasswordHasher using configured params (or defaults)
+        kwargs = {}
+        for key, value in self.config.items():
+            if key in self.hasher_keys:
+                kwargs[key] = value
+
+        import argon2
+        self.password_hasher = argon2.PasswordHasher(**kwargs)
```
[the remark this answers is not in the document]
ehlertjd: Will disable configurability for now.

api/auth/authproviders.py
```diff
@@ -85,6 +88,73 @@ def set_refresh_token_if_exists(self, uid, refresh_token):
...
 
+
+class BasicAuthProvider(AuthProvider):
+    """FOR DEVELOPMENT ONLY!!! Password based authentication.
+
+    This is not production ready until:
+    1. We perform a security audit
+    2. Validate the default parameters
+    3. Provide functionality for a user to change their own password
+    4. Enable password policies (i.e. min length etc)
```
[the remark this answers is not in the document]
ehlertjd: 👍

api/web/base.py
```diff
@@ -255,6 +255,24 @@ def log_in(self):
...
+            self.abort(400, str(e))
+
+        payload = self.request.json_body
+        if 'email' not in payload or 'password' not in payload:
+            self.abort(400, 'email and password required for login')
+
+        token_entry = auth_provider.validate_code(payload)
+        self._generate_session(token_entry)
+
+        return {'token': token_entry['_id']}
```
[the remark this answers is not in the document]
ehlertjd: Once per session, we don't currently invalidate it if we change the password

api/handlers/userhandler.py
ehlertjd: There's a line above that checks for basic auth and aborts the request.

api/auth/authproviders.py
kofalt: Do you think a util func for byte strings and `create_nonce` in turn would call that?

kofalt:
Reduced rigor makes sense to me IFF I see an enormous breaker to throw that prevents this from easily being turned on in production. Stack an "unsafe" feature flag on top of enabling the auth provider?

api/auth/authproviders.py
```diff
@@ -101,6 +101,10 @@ class BasicAuthProvider(AuthProvider):
     def __init__(self):
         super(BasicAuthProvider, self).__init__('basic')
 
+        # Require unsafe_login feature
+        if not config.get_feature('unsafe_login', False):
+            raise NotImplementedError('Unsafe login is not permitted on this site!')
```
kofalt: I was thinking something that explicitly must be set by environment variable (thus requiring a restart of the service and goofing around with yaml), based on my reading of the `get_config` family it looks like this could be set via the database.

Eh, either way is fine.

api/auth/authproviders.py
kofalt: OOB - some improvements to `create_nonce` were suggested, such as dumping urandom directly into a based string, and possible distribution problems with randrange. Something to consider down the road.

In the ideal case, of course, logic such as this is not in our code base at all but is a library call. At the time `create_nonce` was written, my research had only turned up larger frameworks that we could not easily adopt due to our usage of webapp2 et al. Thus this logic was lifted from a library that seemed reasonable under review.

PR #1637, opened 2019-06-20
Es6

hkethi002:
### Review Checklist


ehlertjd:
We should address [FLYW-1071](https://flywheelio.atlassian.net/browse/FLYW-1071) in this PR.

hkethi002:
:+1:, Also, how do we want to document the egg version mismatch that happens when upgrading to es6? Is that something we expect to happen in prod?

PR #1638, opened 2019-06-24
Recursive propogate

roycollins:
Convert the recursive property update to an iterative loop.

On my tests with dev data this reduced the copy session operation from just over 1 minute (63 seconds) to just over 1 second, using the 1234 subject from the dev test database.

I also added indexes on the keys for the jobs.parents collection. We should confirm that we have these indexes, as needed per level, on all the collections.  

db.jobs.ensureIndex({
  "parents.acquisition": 1,
  "parents.session": 1, 
  "parents.subject": 1, 
  "parents.project": 1, 
  "parents.group": 1
})

On my testing the multi key indexes were not being used. I believe we need an array not an object to take advantage of the multi key aspect.  Otherwise, explicitly declaring the indexes above works as expected.  Adding the indexes reduced 8 seconds off the total run time on the tests I ran.  

### Review Checklist


api/dao/containerutil.py
```diff
@@ -51,41 +51,43 @@
 
 NON_OBJECT_ID_COLLECTIONS = ['groups', 'users']
 
-def propagate_changes(cont_name, cont_ids, query, update, include_refs=False):
+def propagate_changes(cont_name, cont_id, query, update, include_refs=False):
```
nagem: I assume we had no calls to this function that were sending lists. I wonder why it was implemented that way...?

api/dao/containerutil.py
```diff
@@ -51,41 +51,43 @@
...
-            config.db[child_cont].update_many(child_query, update)
-            # Recurse to the next hierarchy level
-            propagate_changes(child_cont, child_ids, query, update, include_refs=include_refs)
+    # TODO validate we dont send in invalid data in the update.  Can only be common data to the current level of hierarccy we are updating
+    for cur_cont in containers:
+        config.db[cur_cont].update_many(query, update)
+        if cont_name == cur_cont:
+            return
 
+    raise Exception('Never reached top level container from: {}'.format(cont_name))
```
nagem: The two "Developer Mistake" Exceptions in here add a bumps to the request for that specific type of Exception class. As is, they'll make it user-side if tripped.

api/dao/containerutil.py
roycollins: The only place that is was used was in the recursive calls. The call started from a top level with one container and then passed in all the children in the recursive calls. Now that its all iterative the single Id is all that is necessary.

api/dao/containerutil.py
roycollins: Yes, they will make it user side if they fire.  They should never happen unless we code something incorrectly.  Which is your preference in this case?  I prefer to have it fail hard so that its very unlikely a coding error makes it into the wild.   I could also just log the error and use an exception type that is trapped in the handlers to prevent message from going to clients.  Whichever you prefer I can make that happen, or if you have an alternative approach I can do that too

roycollins:
> Changes look good to me, looks like there is an outstanding TODO to validate a few things via tests.

The TODO is in regard to defining our data model with which data is specifically allowed to be passed into the propagate call.  Just prevents another developer mistake where someone adds in data that can not be pushed all the way down to the bottom of the hierarchy. I will get something in there before this merges.

ehlertjd:
Only thing I'd like to see is let's make sure that the subject propagation is tested and passing from this commit onward (i.e.)

```python
    r = as_admin.get('/subjects/' + subject)
    assert r.ok
    # Move the subject to project 2 
    assert as_admin.put('/subjects/' + subject,
                        json={'project': project2}).ok
    s = as_admin.get('/subjects/' + subject).json()
    print s['parents']
    print s['permissions']
    # Add permission for user on project 2
    uid = as_user.get('/users/self').json()['_id']
    r = as_admin.post('/projects/' + project2 + '/permissions', json={
        '_id': uid,
        'access': 'admin'
        })
    assert r.ok
    proj_perms = as_admin.get('/projects/' + project2).json()['permissions']
    subj_perms =  as_admin.get('/subjects/' + subject).json()['permissions']
    assert subj_perms == proj_perms
```

roycollins:
With the latest merge of master the above tests are not in this branch and passing.  Let me know if there is anything more you see.

PR #1639, opened 2019-06-27
Provider override script

roycollins:
Script to allow overriding site storage provider. Specifically this will create a new provider based on the values in the persistent_fs_url environment variable and set it to be the storage provider on the site container.

### Review Checklist


bin/site_provider_override.py
```diff
@@ -16,39 +17,64 @@
...
+            'bucket': bucket_name,
+            'path': path,
+            'region': params.get('region', None)
+        }
 
-    db.singletons.update({'_id': 'site'}, update)
-    log.info('Providers have been modified')
+        creds = {
+            # we want a key error if these are not set
+            'aws_access_key_id': os.environ['AWS_ACCESS_KEY_ID'],
```
ehlertjd: We'll still have and need to plan for deployments where credentials are managed through the host.

roycollins:
> [#1639 (comment)](https://github.com/flywheel-io/core/pull/1639#discussion_r298736528)
> 
> We'll have to fix this assumption for AWS - we have sites where the access key and secret are not in the environment.

I will need a little context on how we want to address these?  I assume they are using a config file?

PR #1640, opened 2019-06-27
Enable reaper uploads to use group.label instead of group._id [FLYW-1067]

ambrussimon:
Enable `group.label`-based reaper (and label) upload placement.
NOTE: only (reaper) happy path is covered with a test yet!

### Review Checklist


swagger/schemas/definitions/enginemetadata.json
```diff
@@ -35,7 +35,7 @@
                 "group": {
                     "type": "object",
                     "allOf": [{"$ref":"group.json#/definitions/group-metadata-input"}],
-                    "required": ["_id"]
+                    "oneOf": [{"required": ["label"]}, {"required": ["_id"]}]
                 },
```
nagem: @ehlertjd I feel pretty confident in my solo-review but I wanted to ensure this was proper json schema.

ehlertjd: I hadn't seen that before, but have confirmed that it's a valid construct.

tests/integration_tests/python/test_uploads.py
```diff
@@ -270,11 +270,95 @@ def test_reaper_upload(data_builder, randstr, upload_file_form, file_form, as_ad
...
+    }))
+    assert r.ok
+    r = as_admin.post('/resolve', json={'path': [group_4, 'Project Label', 'Session Label4', 'Acquisition Label4']})
+    assert r.ok
+    acq_children = r.json()['children']
+    assert len(acq_children) == 1
+    assert acq_children[0]['name'] == 'group.label.test.4'
+
+
+    # And this group will not be found and create the group_5 tree
```
nagem: This should be "will" instead of "will not"

PR #1642, opened 2019-06-28
(Re-)Add GET /devices/self

ambrussimon:
Devices (eg.: bootstrappers and reapers) would benefit from accessing their own id / api key.

### Review Checklist


tests/integration_tests/python/test_devices.py
```diff
@@ -50,6 +55,15 @@ def test_devices(as_public, as_user, as_admin, as_drone, api_db):
...
     assert 'key' in r.json()
 
+    # try to get device self as user
+    r = as_admin.get('/devices/self')
+    assert r.status_code == 403
+
+    # get device self
+    r = as_drone.get('/devices/self')
+    assert r.ok
+    assert 'key' in r.json()
```
ehlertjd: I'd like to see a little bit more validation on the response body here.

PR #1644, opened 2019-07-01
Get files from elastic by file id [FLYW-1071]

hkethi002:
### Review Checklist


hkethi002:
Yup

nagem:
Automated testing didn't catch this because we mock ES requests and responses. I'd be interested in other's thoughts on where a wider integration test for this should live.

hkethi002:
I'm planning on adding it to the mc integration, which might make sense because the original change that caused this was in the connector

swagger/schemas/definitions/search.json
```diff
@@ -227,7 +227,7 @@
         "search-ml-input-file": {
             "type": "object",
             "properties": {
-                "_id": {"$ref": "common.json#/definitions/objectid"}
+                "_id": {"type": "string"}
```
nagem: There should be one for a file id, if you wanted to use that.

jody-zeitler:
QA verified with the UI branch on the test2 environment

PR #1648, opened 2019-07-09
Group read-only access for users in a project [FLYW-1128]

hkethi002:
### Review Checklist


api/auth/__init__.py
```diff
@@ -42,6 +42,26 @@ def __init__(self, role, privileges):
...
+        get_projects (fn): A function to check if a project has a user's permissions
+
+    Returns:
+        int: The integer value of the user's access to the container
+    """
+    get_access_result = _get_access(uid, group, scope)
+    if get_access_result == -1:
+        if get_projects is not None and len(get_projects()) > 0:
+            # If user has access to any projects of the group, return ro
+            return INTEGER_PERMISSIONS['ro']
```
ehlertjd: Do we need to return `-1` if the `scope` variable is *not* `ro`?

hkethi002: So I think I should change the if statement on L60 to include `and scope is None` because if it is scoped, the group id should be in the scope dictionary, because we shouldn't be doing further checking if the handler is scoped

PR #1651, opened 2019-07-09
Add a generated Go SDK

kofalt:
With help from Justin, generating the Go SDK now works with a few caveats.

1. Struct components are lost on polymorphic returns. This must be corrected down the line.
1. Output is somewhat verbose / wonky / unreadable. This could be improved via custom templates.
1. [Swagger-codegen](https://github.com/swagger-api/swagger-codegen) updated from 2.3 to 2.4.6.
1. Swagger wanted to update / correct its `package-lock.json`, so I let it. This looks to be the right choice - `package.json` is for ranges (`~`, `^`, etc) and lock is for lock.

Overall, this should be considered a "draft" SDK, not eligible to replace our [hand-written](https://github.com/flywheel-io/sdk) work until some further improvements are made and distribution is solved.

### DIFF
--- a/sdk/codegen/src/main/java/io/flywheel/codegen/GoGenerator.java
+++ b/sdk/codegen/src/main/java/io/flywheel/codegen/GoGenerator.java
@@ -1,0 +1,122 @@
+package io.flywheel.codegen;
+
+import io.swagger.codegen.*;
+import io.swagger.codegen.languages.GoClientCodegen;
+import io.swagger.models.ModelImpl;
+import io.swagger.models.Swagger;
+import io.swagger.models.properties.Property;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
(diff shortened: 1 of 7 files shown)


sdk/codegen/src/main/java/io/flywheel/codegen/GoGenerator.java
```diff
@@ -0,0 +1,122 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class GoGenerator extends GoClientCodegen implements CodegenConfig {
+
+    // private static final Pattern SIMPLE_SEMVER_RE = Pattern.compile("(?<semver>\\d+\\.\\d+\\.\\d+)"
+    //         + "(-(?<releaseId>[a-z]+)\\.(?<releaseNum>\\d+))");
```
kofalt: Intentionally leaving some of this stuff in & commented out until we tighten this up later.

PR #1658, opened 2019-07-17
FLYW-733: correct usage of Filename

lubanovic-flywheel:
From the code that it eventually calls (`download()` in `api/handlers/refererhandler.py`), it looks like `Filename` is required but can be empty (swagger barfs if I try to set `required: false`). If empty, it downloads a tar of all files. No regex anywhere here. I updated the YAML text.


roycollins:
The reason swagger wont let you set it not required is because it is a path parameter.  All path parameters are required but they can be empty.  

Good catch on the docs; no regex just a direct equality comparison.

PR #1663, opened 2019-07-25
Fix 66 upgrade to acomodate all possible states of center gears

roycollins:
It seems on new sites the center_gears attribute is set to None instead of []. Reset gears and other scripts seem to set it to the default of []. 

This should handle it is all cases. 

### Review Checklist


ryandavisflywheel:
👍

PR #1664, opened 2019-07-26
Update undelete.py - support subject containers and dry-run

ambrussimon:
Added subject and dry-run support. 
Added bare minimum of tests.
cc @lmperry

### Review Checklist


bin/undelete.py
```diff
@@ -42,7 +42,8 @@ def main(*argv):
     try:
         undelete(args.cont_name, args.cont_id, filename=args.filename,
                  include_parents=args.include_parents,
-                 always_propagate=args.always_propagate)
+                 always_propagate=args.always_propagate,
+                 dry_run=args.dry_run)
```
nagem: OOF

tests/integration_tests/python/test_undelete.py
```diff
@@ -139,5 +139,8 @@ def test_undelete_options(undelete, containers):
     undelete('projects', containers.p_1)
     assert containers.is_deleted('s_1_2')
 
+    undelete('projects', containers.p_1, always_propagate=True, dry_run=True)
+    assert containers.is_deleted('s_1_2')
```
nagem: Shouldn't this be false?

ambrussimon: No, this is rather underhandedly testing `dry_run` instead.

PR #1666, opened 2019-07-26
Api bulk operations 

roycollins:
Creates a bulk API endpoint for performing operations on multiple containers at once.

This PR implements the bulk move for sessions and subjects. As we add more operations and container types we can fill in the remaining methods as needed.

This is an intermediate PR, missing the permissions isolation which are added in the following PR:
https://github.com/flywheel-io/core/pull/1667

### Review Checklist

- Tests were added to cover all code changes
- Is there a DB upgrade or fix?
  - If so is it covered by integeration tests and run against dev?
- Would this PR benefit from a test plan? If so, is one provided?
- Documentation was added / updated
- Code and tests follow standards in CONTRIBUTING.md
- Extra care is taken and log messages are created for [Critical Data](https://github.com/flywheel-io/core/wiki/Critical-Data-Flows)

PR #1667, opened 2019-07-26
Api bulk operations permissions

roycollins:
Adds permission handling for bulk containers and implements the permissions check on the api bulk endpoint.

### Review Checklist


api/auth/containerauth.py
```diff
@@ -199,3 +199,48 @@ def f(method, query=None, user=None, public=False, projection=None, pagination=N
             query['public'] = True
         return exec_op(method, query=query, user=user, public=public, projection=projection, pagination=pagination)
     return f
+
+
+def validate_container_permissions(containers, container_type, user_id, permission):
+
```
ehlertjd: minor - conventionally, no space between method declaration and docstring.

api/auth/containerauth.py
```diff
@@ -199,3 +199,48 @@ def f(method, query=None, user=None, public=False, projection=None, pagination=N
...
+    Arguments:
+        containers (list): The list containing Ids of containers to check
+        container_type (string): The container type we are checking
+        user_id (str): id of the user to check
+        permission (str): Permission we would like to validate; ro, rw, admin
+
+    Returns:
+        bool: True if access is permitted, false otherwise
+    """
```
ehlertjd: 👌

api/auth/containerauth.py
```diff
@@ -199,3 +199,48 @@ def f(method, query=None, user=None, public=False, projection=None, pagination=N
...
+
+    permission_levels = ['admin', 'rw', 'ro']
+    container_types = ['acquisitions', 'sessions', 'subjects', 'projects', 'groups']
```
ehlertjd: Do we have these declared anywhere else that we can leverage here?
Use tuples rather than lists as they are immutable.

api/dao/containerstorage.py
```diff
@@ -392,23 +396,47 @@ def move_sessions_to_project(self, source_list, dest_project, conflict_mode=None
...
+                {'$match': {
+                    'subject_doc.code': {'$in': conflict_subject_codes}
+                }}
+            ])
+
+            conflict_session_ids = []
+            for conflict in conflict_sessions:
+                conflict_session_ids.append(conflict['_id'])
+
+            return conflict_session_ids
```
ehlertjd: You could use a list comprehension here:
```python
return [conflict['_id'] for conflict in conflict_seessions]
```

api/data_views/util.py
```diff
@@ -123,6 +123,8 @@ def convert_to_datatype(value, datatype):
 
     except ValueError:
         return nil_value
+    except TypeError:
+        return nil_value
```
ehlertjd: Is this an extraneous merge patch?

api/auth/containerauth.py
roycollins: They are converted to tuples over lists but I need them to be in that exact order.  If I pull them from the top level permissions then anyone adding a permission could potential change the order and break it unintentionally.  I would like to make them all constants in the future.

roycollins:
All the comments from the prior PR #1666 have been addressed, expect the API name. Everything is now in this branch including the recently merges commits needed from the 2 recursive propagation and permission fix branches.  Is now ready for a final review.

api/dao/containerstorage.py
```diff
@@ -339,6 +339,255 @@ def get_all_for_targets(self, target_type, target_ids, user=None, projection=Non
...
+                move_subjects.append(session['_id'])
+                continue
+            # If we are moving all the sessions then we dont need to copy the subject.
+            if set(session['sessions']).issubset(source_list_set):
+                move_subjects.append(session['_id'])
+                continue
+            copy_subjects.append(session['_id'])
+
+        # quick sanity check
+        assert len(source_subject_ids) == (
```
ehlertjd: What action should be taken if this assertion fails? Can this be a check/log abort request instead? That might make troubleshooting easier.

api/dao/containerutil.py
```diff
@@ -60,35 +62,77 @@ def propagate_changes(cont_name, cont_id, query, update, include_refs=False):
...
+
+    # Non standard containers only need to update related analysis and jobs, if any
+    if cont_name not in CONTAINER_PROPAGATE:
+        return
+
+    # TODO validate we dont send in invalid data in the update.  Can only be common data to the current level of hierarccy we are updating
+    for cur_cont in CONTAINER_PROPAGATE:
+        if cont_name == cur_cont:
+            for key in ['parents.group', 'parents.project', 'parents.subject', 'parents.session', 'parents']:
+                if query.get(key):
```
ehlertjd: These two lines could just be `query.pop(key, None)`

bin/oneoffs/migrate_storage.py
```diff
@@ -116,10 +115,10 @@ def get_containers_files(containers_prefixes):
                     'fileinfo': f,
                     'prefix': prefix
                 }
-                # If filter_source (global) is set to a provider_id, only
-                # append files that have the same BSON ID
-                if filter_source and str(f['provider_id']) != filter_source:
+
+                if filter_source and f.provider_id != filter_source:
```
ehlertjd: I think this shouldn't be in the diff, should it?

swagger/schemas/definitions/bulk.json
```diff
@@ -0,0 +1,39 @@
...
+                    "items": {
+                        "type": "string"
+                    },
+                    "description": "Array with a single continer Id of the destination container's Id"
+                },
+                "conflict_mode": {
+                    "type": "string",
+                    "items": {
+                        "type": "string",
+                        "enum": ["skip", "copy", null]
```
ehlertjd: The example has conflict_mode of `move`?

roycollins: It should be move or skip.  Fixed.

bin/oneoffs/migrate_storage.py
roycollins: Yeah should not have been in there.  Fixed

api/handlers/bulkhandler.py
```diff
@@ -0,0 +1,128 @@
...
+    		ii. Or skip the move entirely.
+        '''
+
+        if (self.payload.get('conflict_mode') is None or self.payload['conflict_mode'] == ''):
+            return self.source_storage.move_sessions_to_project(self.source_list, self.dest_list[0], conflict_mode=None)
+
+        self.source_storage.move_sessions_to_project(self.source_list, self.dest_list[0], conflict_mode=self.payload.get('conflict_mode'))
+
+
+    def _move_sessions_to_subjects(self):
```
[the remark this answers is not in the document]
roycollins: Yeah we have a conflict mode for move_sessions_to_project but we dont have a conflict_mode for move_sessions_to_subject since they cant have conflicts.  I suppose we should have add one more level to the api so that we could have unique json schemas per container/operation/destination.  
As it is I made it return nothing on the dry run since there will never be conflicts but now you have to explicitly supply a conflict mode of 'skip' or 'move' to make the operation actually go.  

Im not sure that is more or less confusing. Since they are unique operations with unique options we should really add another level of distinction for each move type destination.

api/dao/containerutil.py
roycollins: Nice.  Good pointer.  Fixed

api/dao/containerstorage.py
roycollins: The only way this should ever fire is if the queries above were changed or the arrays were mutated prior to this check.  Its more just a programmer notice in the code that these 3 values should be a full set.  Asserts are generally used for cases that should never happen.  I assume the log will have the line that the assert failed.  But I can add in an additional log if you think that can be used.

PR #1669, opened 2019-07-29
Remove int check on database version only for upgrade [FLYW-1149]

roycollins:
Removes the explicit int check on the upgrade. 
Casts to int so floats will work as expected but even strings with numbers will work.

### Review Checklist


bin/database.py
```diff
@@ -62,7 +62,7 @@ def confirm_schema_match():
     """
 
     db_version, applied_fixes, applied_checks = get_db_version()
-    if not isinstance(db_version, int) or db_version > CURRENT_DATABASE_VERSION:
+    if not db_version or db_version > CURRENT_DATABASE_VERSION:
```
ehlertjd: I missed this on the first pass, I think it should be if `db_version is None`. Right now we error out on initial install.

PR #1670, opened 2019-07-29
Check for attempt if attempt_n is not in the job map [FLYW-1190]

hkethi002:
Prevent retry orphan loops
### Review Checklist


api/jobs/queue.py
```diff
@@ -240,7 +240,7 @@ def enqueue_job(job_map, origin, perm_check_uid=None):
 
         # Add job tags, config, attempt number, and/or previous job ID, if present
         tags            = job_map.get('tags', [])
-        attempt_n       = job_map.get('attempt_n', 1)
+        attempt_n       = job_map.get('attempt_n', job_map.get('attempt', 1))
```
ehlertjd: This is the only use of `attempt_n` that I see, can this line just be `attempt = job_map.get('attempt', 1)`?

PR #1672, opened 2019-08-01
Allow insecure http for SDK create_drone_client [FLYW-1259]

ehlertjd:
Updated usage:
```
fw = flywheel.create_drone_client(..., port=8080, _force_insecure=True)
```

### Review Checklist


sdk/src/python/tests/integration_tests/sdk_test_case.py
```diff
@@ -32,12 +32,42 @@ def get_api_key():
 
     return api_key, device_key
 
+def make_drone_client():
+    if init_db.SCITRAN_PERSISTENT_DB_URI:
```
roycollins: I assumed this has to be set for the tests to run. Im just curious when this is not true.

ehlertjd: I will occasionally run the SDK tests directly against my running local dev environment by setting the `SdkTestKey` to an api-key, which is when this clause is inverted. I'm sure that we could devise a better way to do low-latency iterations given sufficient time and motivation.

sdk/src/python/tests/integration_tests/sdk_test_case.py
```diff
@@ -32,12 +32,42 @@ def get_api_key():
...
+
+
 def make_clients():
     api_key, device_key = get_api_key()
 
     fw = flywheel.Flywheel(api_key)
     if device_key:
         fw_device = flywheel.Flywheel(device_key)
+    else:
+        fw_device = None
```
hkethi002: Is there a reason we're using device and drone, I know that we use them interchangeably in core, but this maybe a larger topic of keeping names and variables consistent.

sdk/codegen/src/main/resources/fw-python/drone_login.py
```diff
@@ -20,7 +20,11 @@ def create_drone_client(host, secret, method, name, port=443, **kwargs):
         util.set_verify_ssl(session)
 
         # Get auth status to determine device id
-        base_uri = 'https://{}:{}/api'.format(host, port)
+        force_insecure = kwargs.pop('_force_insecure', False)
+        if force_insecure:
```
ryandavisflywheel: `assert(type(force_insecure) == bool)` 
here so it blows up for anything else? ... maybe I'm being paranoid, but a
`force_insecure = "False"` 
among others vals/types for force_insecure are truthy

sdk/src/python/tests/integration_tests/sdk_test_case.py
ehlertjd: Good point - I'm going to leave it for now, but maybe a refactor would be in order if we ever start making the distinction clearer.

PR #1676, opened 2019-08-12
Allow auto-update of rules for invalid gears [FLYW-1315]

ehlertjd:
### Review Checklist

### DIFF
--- a/api/jobs/gears.py
+++ b/api/jobs/gears.py
@@ -64,11 +64,8 @@ def get_gear(_id):
         raise APINotFoundException('Cannot find gear {}'.format(_id))
     return gear
 
-def get_latest_gear(name, include_invalid=False):
-    query = {'gear.name': name}
-    if not include_invalid:
-        query['gear.custom.flywheel.invalid'] = {'$ne': True}
-    gears = config.db.gears.find(query).sort('created', direction=-1).limit(1)
+def get_latest_gear(name):
+    gears = config.db.gears.find({'gear.name': name, 'gear.custom.flywheel.invalid': {'$ne': True}}).sort('created', direction=-1).limit(1)
     if gears.count() > 0:
         return gears[0]
 
@@ -183,7 +180,7 @@ def filter_optional_inputs(geardoc):
 
 def insert_gear(doc):
     gear_tools.validate_manifest(doc['gear'])
-    last_gear = get_latest_gear(doc['gear']['name'], include_invalid=True)
+    installed_gears = get_all_gear_versions(doc['gear']['name'])
 
     # This can be mongo-escaped and re-used later
     if doc.get("invocation-schema"):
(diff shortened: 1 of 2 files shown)


nagem:
With this solution you can't "roll back" a gear, so that is something to note. For example:
  - Gear rules are on gear v.1 with auto-update turned on
  - Gear v.2 rolls out, gear rules are updated
  - Gear v.2 is deemed problematic and disabled
  - Gear rules are manually brought back to Gear v.1
  - Gear v.3 rolls out, gear rules are not updated as the find will return Gear v.2

nagem:
Can the logic be changed to look for gear rules with a specific gear alg name or is that not stored on the gear rule?

ehlertjd:
@nagem I've adjusted the logic to update *any* rule on a prior gear version (with auto-update enabled), not just most-recent.

tests/integration_tests/python/test_rules.py
```diff
@@ -859,6 +859,78 @@ def test_auto_update_rules(data_builder, api_db, as_admin):
     assert r.ok
 
 
+def test_auto_update_invalid_rule(data_builder, api_db, as_admin):
```
ryandavisflywheel: Unit test!  ... very nice.  This makes it easier to understand this PR's aim.

tests/integration_tests/python/test_rules.py
```diff
@@ -859,6 +859,78 @@ def test_auto_update_rules(data_builder, api_db, as_admin):
...
+        'any': [],
+        'not': [],
+        'all': [{'type': 'file.type', 'value': 'tabular data'}],
+        'disabled': False,
+        'auto_update': True
+    })
+    assert r.ok
+    rule2_id = r.json()['_id']
+
+    # Should auto-update all rules to gearv2
```
ryandavisflywheel: This comment's assertion doesn't actually gets asserted... maybe a GET rules and assert gearv2?

api/jobs/gears.py
```diff
@@ -222,8 +223,8 @@ def check_for_gear_insertion(doc):
     if conflict is not None:
         raise Exception('Gear "' + doc['gear']['name'] + '" version "' + doc['gear']['version'] + '" already exists, consider changing the version string.')
 
-def auto_update_rules(gear_id, last_gear_id):
-    config.db.project_rules.update_many({'gear_id': str(last_gear_id), 'auto_update': True}, {"$set": {'gear_id': str(gear_id)}})
+def auto_update_rules(gear_id, installed_gear_ids):
+    config.db.project_rules.update_many({'gear_id': {'$in': installed_gear_ids}, 'auto_update': True}, {"$set": {'gear_id': str(gear_id)}})
```
ryandavisflywheel: ... some new MongoDB syntax to me here... cool :)

PR #1677, opened 2019-08-12
Allow inclusion of invalid when listing gears [FLYW-1321]

ehlertjd:
This will be necessary for upgrades of some gears via CLI.

### Review Checklist

### DIFF
--- a/api/jobs/gears.py
+++ b/api/jobs/gears.py
@@ -20,16 +20,19 @@ from ..web.errors import APIValidationException, APINotFoundException
 
 log = config.log
 
-def get_gears(all_versions=False, pagination=None):
+def get_gears(all_versions=False, pagination=None, include_invalid=False):
     """
     Fetch the install-global gears from the database
     """
+    match = {}
+    if not include_invalid:
+        match['gear.custom.flywheel.invalid'] = {'$ne': True}
 
     if all_versions:
         kwargs = {
             # Invalid disables a gear from running entirely.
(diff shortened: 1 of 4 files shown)


api/jobs/gears.py
```diff
@@ -20,16 +20,19 @@
 
 log = config.log
 
-def get_gears(all_versions=False, pagination=None):
+def get_gears(all_versions=False, pagination=None, include_invalid=False):
     """
     Fetch the install-global gears from the database
     """
+    match = {}
+    if not include_invalid:
```
ryandavisflywheel: I've brought this up in the past and realize that the issue lies in several layers here (invalid is in the DB), but a good rule of thumb I tend to follow is positive naming of variables i.e. `include_valid_only=True` vs `include_invalid=False` which makes our logic easier to understand by avoiding the use of double-negatives thus removing extra unnecessary mental gymnastics.
`if include_valid_only:` is 1 less logical step than `if not include_invalid` and when we sit and look at code all day these things add up quickly.

nagem: Let me point you to a Jira ticket about that very subject!

https://flywheelio.atlassian.net/browse/FLYW-1314

It also has a bit of history about that key, specifically that it was never meant to be permanent.

nagem: I removed my comment because I thought were were talking about the key on the gear. That fix is part of this ticket:

https://flywheelio.atlassian.net/browse/FLYW-1314

PR #1680, opened 2019-08-16
Increment document revision when updating records [FLYW-1283]

ehlertjd:
### Review Checklist


swagger/schemas/definitions/common.json
```diff
@@ -90,6 +90,13 @@
                     "type": "boolean"
                 }
             }
+        },
+        "revision": {
+            "oneOf": [
+                {"type": "integer"},
+                {"type": "null"}
```
roycollins: We seem to always set the revision to 1.  If you dont use this definition on the ingress API filter then we can guarantee that we always have a revision number.  Would just be a safe check that our internal model is always consistent.  Outbound would always have the revision as well.

ehlertjd: We can change this here to always be type integer, but that wouldn't address the presence of the value (i.e. as in making it required).
I'm happy to update the type, but hesitant to make this a hard required field right now - given the complexity of core it seems completely feasible to have missed a spot.

api/dao/containerstorage.py
```diff
@@ -177,7 +177,6 @@ def _from_mongo(self, cont):
 
     def create_or_update_el(self, payload, origin, **kwargs):
         if self.dbc.find_one({'_id': payload['_id'], 'deleted': {'$exists': False}}):
-            payload['modified'] = datetime.datetime.utcnow()
```
hkethi002: Should we be calling `update_modified_and_revision` on the payload here?

PR #1681, opened 2019-08-16
Allow site storage when multiproject is off [FLYW-1351]

roycollins:
When multiproject is disabled always return the site storage provider 

### Review Checklist


api/site/storage_provider_service.py
```diff
@@ -26,6 +26,9 @@ def determine_provider(self, origin, container=None, file_size=None, force_site_
         if not site_doc.providers.get('storage'):
             raise ValueError('Site settings are not configured for a storage provider')
 
+        if not config.is_multiproject_enabled():
```
ehlertjd: Explicit is better than implicit, but also DRY --- this could be an additional clause to a single `if (should use site provider) then (use site provider)` statement, currently we have 3 of those.

roycollins: I prefer explicit in this case so that we can have the guard conditions right at the top so you completely negate the service logic if the preconditions are not met.  If we do it in the logic below then you have to drill down into the decision logic to see what the preconditions are for the function.

PR #1682, opened 2019-08-21
Set Lab edition to False by default [FLYW-1345]

roycollins:
Change the default state for new groups and projects to have Lab Edition False by default.
The tests are still going to use True so they dont run into ad-hoc restrictions.

### Review Checklist

  - If so is it covered by integration tests and run against dev?


tests/integration_tests/python/test_access_log.py
```diff
@@ -517,11 +517,8 @@ def test_job_access(data_builder, as_admin, as_drone, log_db, default_payload,
     }
     gear = data_builder.create_gear(gear=gear_doc)
 
-    project = data_builder.create_project()
-    # Projects must have a provider for gear uploads to work 
-    update = {'providers': {'storage': 'deadbeefdeadbeefdeadbeef'}}
-    r = as_admin.put('/projects/' + project, json=update)
-    assert r.ok
+    # Projects must have a provider for gear uploads to work
+    project = data_builder.create_project(providers={'storage': 'deadbeefdeadbeefdeadbeef'})
```
ryandavisflywheel: Nice... I saw a lot of opportunities to simplify in this way after doing some reverse-engineering of data_builder's implicit ways

api/dao/containerstorage.py
```diff
@@ -30,15 +30,15 @@ def __init__(self):
 
     def _to_mongo(self, payload):
         # Ascertain defaults in our model for consistency
-        payload.setdefault('editions', {}).setdefault('lab', True)
```
ryandavisflywheel: My gut tells me we should be doing this sort of logic (setting defaults) at a lower level with schema so we can be validating changes via unit tests... my only criticism

PR #1683, opened 2019-08-21
Fix bug where dot notation access to dict in migration storage fails if source arg is used

ryandavisflywheel:


tests/integration_tests/python/test_migrate_storage.py
```diff
@@ -125,15 +124,12 @@ def files_to_migrate(data_builder, api_db, as_admin, randstr, file_form):
             pass
 
 
-
-
-def test_migrate_containers(files_to_migrate, as_admin, migrate_storage, second_storage_provider):
+def test_migrate_containers(files_to_migrate, as_admin, migrate_storage, second_storage_provider, api_db):
```
davidfarkas: One thing, looks like `api_db` is not used in this test.

ryandavisflywheel: Ah, good catch.  It's been removed :)

PR #1687, opened 2019-08-30
Add one off fix for missing input id [FLYW-1473]

roycollins:
Adds a one -off check for the missing provider id found in production data.

### Review Checklist


bin/database.py
```diff
@@ -2599,6 +2604,43 @@ def upgrade_provider_id(storage_id):
...
+    on an analyses file and these are the steps we can take to located that
+    analyses file. All other cases will most likely be unique and we can
+    add them as we come across them
+    """
+    if collection == 'analyses':
+        saved_ids = []
+        for a in bad_files:
+            made_updates = False
+            for i in a['inputs']:
+                if not i.get('_id'):
```
ryandavisflywheel: a more pythonic way is `if i.get('_id') is None` ... minor nitpick

bin/database.py
```diff
@@ -2599,6 +2604,43 @@ def upgrade_provider_id(storage_id):
...
+                        if more_related and more_related_type == 'acquisitions':
+                            for f in more_related['files']:
+                                if f.get('hash', 'wont') == i.get('hash', 'match'):
+                                    i['_id'] = f['_id']
+                                    made_updates = True
+            if made_updates:
+                config.db.analyses.update({'_id': a['_id']}, {'$set': {'inputs': a['inputs']}})
+                saved_ids.append(a['_id'])
+        if saved_ids:
+            print 'Alhamdulillah!  We actually fixed some {}, Precisely {}'.format(
```
ryandavisflywheel: :rolling_on_the_floor_laughing:

bin/database.py
roycollins: I like pythonic,  thanks.

PR #1688, opened 2019-08-30
fix permission recursion [FLYW-1475]

roycollins:
The permissions are not recursing on all the containers. The lookup is done via the <parent_name> attribute and that is not always set or may be a string vs objectId. That causes the lookup to miss the children. 
Replacing the lookup with <parents.parent_name> solves the issue.

I added tests that specifically isolate the conditions in which this was found.

## Review Checklist


api/dao/containerutil.py
```diff
@@ -77,7 +77,9 @@ def propagate_changes(cont_name, cont_ids, query, update, include_refs=False):
 
     if cont_name in containers[:-1]:
         child_cont = containers[containers.index(cont_name) + 1]
-        child_ids = [c['_id'] for c in config.db[child_cont].find({singularize(cont_name): {'$in': cont_ids}}, [])]
+        child_attr = 'parents.{}'.format(singularize(cont_name))
+        child_ids = [c['_id'] for c in config.db[child_cont].find(
+            {child_attr: {'$in': cont_ids}}, [])]
```
ehlertjd: Was this defect introduced prior to 10.0? If so, we'll likely need a database fix

roycollins: That code was last changed on Dec 7 2017 but I am not sure what the version was.  Way prior to 10.0 either way.  

I think the best would be to propagate permissions on all projects since the permissions should be the same for all sub containers. Is this correct?

ehlertjd: If it was only affecting the subjects collection, I'd prefer to limit the number of updates we try to do during the update, but effectively yeah we'd want to re-push every project to its subjects.

PR #1689, opened 2019-08-30
Subject jobs endpoint [FLYW-1033]

roycollins:
Add jobs to the subject endpoint.
Allow filtering on the endpoint to remove inputs that belong to a project which the user does not have permission to read.  

This does require changes in #1688 before it will pass all tests as permissions are not set correctly on child containers which are now used in the filtering

### Review Checklist


roycollins:
The endpoint has been converted to project level now with filters.  Can you just sanity check this to be sure it seems to be on the right track.

ehlertjd:
I'm fairly certain we decided that we didn't need to / wasn't scalable to pull related containers, rather relying on filtering by the job's `parents` or `destination` instead. I think this would greatly simplify the logic as we'd have a single collection retrieval and can leverage the existing pagination.

roycollins:
Closed for a different appraoch

PR #1690, opened 2019-09-03
add description and origin to dataviews [FLYW-1182]

roycollins:
Allow description to be supplied on dataviews and storage origin with dataviews on creation.

### Review Checklist


swagger/schemas/definitions/origin.json
```diff
@@ -0,0 +1,18 @@
+{
```
ehlertjd: We should probably further consolidate 'origin', I believe it is used in multiple places.
[FLYW-1479](https://flywheelio.atlassian.net/browse/FLYW-1479) to track.

swagger/schemas/definitions/provider.json
```diff
@@ -41,7 +28,7 @@
                     "type": "string",
                     "description": "A human readable label for the provider"
                 },
-                "origin": {"$ref": "#/definitions/provider-origin"},
+                "origin": {"$ref": "origin.json#/definitions/origin"},
```
ehlertjd: 👌

swagger/schemas/definitions/origin.json
roycollins: Good call.  I knew of the provider one so I threw it in there. The others look similar but I am not sure they are the same.   Good for a different PR, agreed.

PR #1693, opened 2019-09-10
Anti virus check [FLYW-1364]

davidfarkas:
Add Anti-Virus check feature. In high-level:
- put user uploaded files into quarantine state.
- send a webhook to a specific endpoint about the newly uploaded file
  - the webhook payload contains a signed download url and a signed response url
- callback handler to receive the result of the virus scanning

For more details see the Jira ticket.

Required PR before this merges:
https://github.com/flywheel-io/storage-plugins/pull/55

### Review Checklist


api/signed_urls.py
```diff
@@ -0,0 +1,44 @@
...
+import hmac
+import time
+import urlparse
+from urllib import quote, urlencode
+
+from . import config
+from .web import errors
+
+
+def generate_signed_url(url, method='GET', expires_in=3600):
```
ehlertjd: Can we get some documentation on how this algorithm was chosen and what security considerations were given?

api/handlers/listhandler.py
```diff
@@ -424,19 +425,28 @@ def get(self, cont_name, list_name, **kwargs):
...
+        elif signature:
+            verify_signed_url(self.request.url, 'GET')
+            permchecker = always_ok
 
         # Grab fileinfo from db
         fileinfo = keycheck(permchecker(storage.exec_op))('GET', _id, query_params=kwargs)
         if not fileinfo:
             self.abort(404, 'no such file')
 
+        if config.get_feature('virus_scan'):
```
ehlertjd: Consistently pass `False` for the default here.

api/placer.py
```diff
@@ -90,6 +91,23 @@ def requireMetadata(self):
         if self.metadata == None:
             raise FileFormException('Metadata required')
 
+    def before_save_file(self, file_attrs):
+        if not config.get_feature('virus_scan', False):
+            return
+        if self.origin['type'] == 'user':
+            callback_url = config.get_config()['webhooks']['virus_scan']
+            if not callback_url:
```
ehlertjd: Add a crit log statement if this happens.

api/placer.py
```diff
@@ -90,6 +91,23 @@ def requireMetadata(self):
...
+                raise APIException('Callback url for virus scan webhook is not configured')
+            webhook = VirusScanWebhook(callback_url)
+            failuers = webhook.call(file_info=file_attrs, parent={
```
ehlertjd: `failuers` -> `failures`

api/webhooks/base.py
```diff
@@ -0,0 +1,26 @@
...
+    def __init__(self, callback_urls):
+        self.session = Session()
+        self.callback_urls = callback_urls
+
+    def build_request_body(self, *args, **kwargs):
+        raise NotImplementedError
+
+    def call(self, *args, **kwargs):
+        raise_for_status = kwargs.pop('raise_for_status', False)
+        failuers = []
```
ehlertjd: `failuers` -> `failures`

api/webhooks/base.py
```diff
@@ -0,0 +1,26 @@
...
+        for url in self.callback_urls:
+            payload = self.build_request_body(*args, **kwargs)
+            r = self.session.post(url, data=json.dumps(payload, default=custom_json_serializer))
```
ehlertjd: We should also log any failures at the error level.

api/webhooks/base.py
```diff
@@ -0,0 +1,26 @@
+import json
+
+from requests import Session
+
+from ..web.encoder import custom_json_serializer
+
+
+class BaseWebhook(object):
+    def __init__(self, callback_urls):
+        self.session = Session()
```
ehlertjd: I don't know what the default timeout is for `requests` but I suspect that it is far too lenient for service to service communication - can we set it to a reasonable default (maybe 5 seconds or so?)

api/signed_urls.py
davidfarkas: I added some documentation. Let me know if something missing, or need more information.

api/webhooks/base.py
davidfarkas: Good point. FYI according to the doc: `By default, requests do not time out unless a timeout value is set explicitly. Without a timeout, your code may hang for minutes or more.`

davidfarkas:
> Overall this is great work. I would like to see additional logging & documentation before it merges.
> 
> Who resends webhooks in the event of a failure - do we need some process for that?

It's a good question. The virus scanning spec says the following:

```
If the service is unavailable, the upload will fail to persist
 - If this is unacceptable, a “cleanup” sweep can be scheduled that attempts to resend all quarantined files’ information to the Genentech service
 - Monitoring infrastructure within Flywheel will be made aware of any service outages
```

@nagem Any thoughts on this?

tests/unit_tests/python/test_signed_urls.py
```diff
@@ -0,0 +1,41 @@
...
+    with pytest.raises(errors.APIPermissionException) as error:
+        signed_urls.verify_signed_url(signed_url, 'GET')
+    assert 'Expired signed url' in str(error)
+    # manipulated url -> invalid signature
+    url_parts = list(urlparse.urlparse(signed_url))
+    query = dict(urlparse.parse_qsl(url_parts[4]))
+    query['expires'] = 3900
+    url_parts[4] = urlencode(query)
+    url = urlparse.urlunparse(url_parts)
+    with pytest.raises(errors.APIPermissionException) as error:
```
ehlertjd: Also, even though it appears to be correct - please explicitly test changing just the path & method portions of the URL.

tests/unit_tests/python/test_uploads.py
```diff
@@ -217,5 +224,61 @@ def test_signed_url_filelisthandler_upload(as_drone, data_builder, mocker):
...
+
+    # upload file as user
+    r = as_user.post('/acquisitions/' + acquisition + '/files', POST=file_form(
+        file_name, meta={'name': file_name, 'type': 'csv'}))
+    assert r.ok
+    # user uploaded file is quarantined
+    r = as_user.get('/acquisitions/' + acquisition + '/files/test.csv')
+    assert not r.ok
+    assert r.status_code == 400
+
```
ehlertjd: We should also test that file is accessible via the signed url provided to the virus scan webhook.

nagem:
@davidfarkas I'll add two new tickets so this PR can get resolved after the rules triggering logic is added and tested. 

The two new tickets:
  - https://flywheelio.atlassian.net/browse/FLYW-1541
    - Add a Prometheus metrics for file states and counts
  - https://flywheelio.atlassian.net/browse/FLYW-1542
    - Add method to resubmit all quarantined files for scanning

@ehlertjd or @davidfarkas let me know if I missed anything in those additional tasks

davidfarkas:
> One thing that appears to be missing is _not_ triggering rules if a file goes into quarantine.

Done and extended `test_upload_with_virus_scan_enabled` to test this.

api/signed_urls.py
```diff
@@ -38,15 +38,16 @@ def generate_signed_url(url, method='GET', expires_in=3600):
...
+    parsed_url = urlparse.urlparse(url)
+    parsed_url = parsed_url._replace(path=quote(parsed_url.path))  # parsed_url is a namedtuple
+    host_name = parsed_url.netloc.rsplit(':', 1)[0]
+    signature = _calc_hash(method, host_name, parsed_url.path, expires)
+    query = dict(urlparse.parse_qsl(parsed_url.query))
     query['expires'] = expires
-    query['signature'] = signature
-    url_parts[4] = urlencode(query)
-    return urlparse.urlunparse(url_parts)
+    query['signature'] = base64.urlsafe_b64encode(signature)
```
ehlertjd: 👌

api/jobs/rules.py
```diff
@@ -211,6 +211,12 @@ def create_potential_jobs(db, container, container_type, file_, rule_failure_cal
 
     potential_jobs = []
 
+    # Skip if virus scan feature enabled and the file is quarantined
+    if config.get_feature('virus_scan', False):
+        virus_scan_state = file_.get('virus_scan', {}).get('state')
+        if virus_scan_state and virus_scan_state != 'clean':
```
ehlertjd: Add an info statement - "Skipping rule evaluation for {container_type}: {container_id} {filename} - file is not clean"

api/webhooks/base.py
```diff
@@ -0,0 +1,31 @@
...
+class BaseWebhook(object):
+    def __init__(self, callback_urls, request_timeout=5):
+        self.session = Session()
+        self.callback_urls = callback_urls
+        self._request_timeout = request_timeout
+
+    def build_request_body(self, *args, **kwargs):
+        raise NotImplementedError
+
+    def call(self, *args, **kwargs):
```
ambrussimon: Too bad py2 cannot yet handle `*args` _and_ `raise_for_status=False`...

api/webhooks/base.py
```diff
@@ -0,0 +1,31 @@
...
+        for url in self.callback_urls:
+            payload = self.build_request_body(*args, **kwargs)
+            try:
+                r = self.session.post(url, data=json.dumps(payload, default=custom_json_serializer), timeout=self._request_timeout)
+                r.raise_for_status()
+            except exceptions.RequestException as e:
+                config.log.error('Webhook failed for the following callback url: %s, details: %s', url, e)
+                failures.append(e)
+                if raise_for_status:
+                    raise
```
ambrussimon: When would we actually want this raised?

api/webhooks/base.py
```diff
@@ -0,0 +1,31 @@
+import json
+
+from requests import Session, exceptions
+
+from .. import config
+from ..web.encoder import custom_json_serializer
+
+
+class BaseWebhook(object):
```
ambrussimon: The `BaseWebhook` abstraction seems a little early TBH, but is nevertheless the right direction.

api/webhooks/base.py
davidfarkas: discussed offline, I use this when resend failed webhooks, open PR soon

PR #1694, opened 2019-09-11
Jobs filter [FLYW-1526]

roycollins:
Creates a jobs endpoint on project that will process all jobs on that project level or lower. 
Converts pagination filters to be used in the parents dict on the mongo collection. This conversion should move all filter checking to the higher parents level rather than on the lower level named attributes, such as 'project' or 'subject'.  The tests all pass with this change but I do want to discuss it to be sure this is both sound and that the test suite looks for this type of validation.

Also have one outstanding question on the permchecker as it seems to allow users without read access so a little pointer on how I am misconstructing that would be helpful. 

### Review Checklist


api/dao/dbutil.py
```diff
@@ -80,6 +81,11 @@ def paginate_find(collection, find_kwargs, pagination):
             filter_.update(pagination['filter'])
             find_kwargs['filter'] = filter_
 
+            (filter_, parents_query) = _append_parents(filter_)
```
ehlertjd: I think this would be cleaner (and easier to understand) if we took an optional `query` parameter to this function, and applied it as an update to filter after processing everything else.

roycollins: I did that originally but I perfered keeping the focus to be finding and removing the parent containers from the filter dictionary.  The function itself should not know or care what the result is going to be used for so its up to the caller to arrange the pieces after it gets them back.  
If anything we can wrap this into two functions that put the parent results in 2 different places depending on which is calling it. If we take this approach then in the future if there is a third placement arrangement we only have to build a new composition function calling our current function and none of the underlying unit tests for the base function will need to be changed at all.   We will just have to build unit tests for the new composite function.

I just noticed the function is still called 'append_parents', as my first approach dictated but I will get that renamed to filter_parents or remove_parents.   

Let me know your thoughts on this approach.  I usually skip the composition until the third time something is used, currently its only2, just to avoid the overhead involved when it might never get used again.  If we do have a third type in the future we can add the wrapper at that time.

ehlertjd: As discussed offline:
1. Use a consistent function signature (i.e. filters should be in the 'filter' property of kwargs)
2. Ensure that any parent filter provided in kwargs "wins" (e.g. client can't escape the intended filter by specifying a different project)

roycollins:
Converted the signature and added override for find_kwargs.

api/dao/dbutil.py
```diff
@@ -68,6 +68,9 @@ def paginate_find(collection, find_kwargs, pagination):
     Raises PaginationError if the query is incompatible with the pagination:
      * `sort` in find_kwargs and `after_id` in pagination
     """
+
+    find_kwargs = _append_parents(find_kwargs)
```
ehlertjd: Shouldn't this be `_append_parents(find_kwargs.get('filter', {}))`?

roycollins: Good catch.  It must not have been used in previous integration tests. I added a unit test to paginate_find to cover at least the cases above to prevent them from breaking in the future.  Its not fully covered but a step closer now.

PR #1696, opened 2019-09-17
Validate jobs are setting required inputs [FLYW-1533]

hkethi002:
### Review Checklist


tests/integration_tests/python/test_batch.py
```diff
@@ -463,6 +470,7 @@ def test_no_input_context_batch(data_builder, default_payload, as_admin, file_fo
     gear_name = randstr()
     gear_doc = default_payload['gear']
     gear_doc['gear']['name'] = gear_name
+    # gear_doc['gear']['inputs']['test_context_value'] = {'base': 'context'}
```
nagem: Is this a residual from testing?

hkethi002: yup

api/jobs/job_util.py
```diff
@@ -2,16 +2,48 @@
...
+
+    Args:
+        job_map (dict): A job object to validate
+        gear_doc (dict): The gear object of the job
+
+    Raises:
+        InputValidationException: raised if invalid job
+    """
+    # Ensure that all inputs in the job are valid and that all required input
+    job_map = copy.deepcopy(job_map)
```
ehlertjd: Do we need this copy? I don't see where we're modifying `job_map` or its contents

tests/unit_tests/python/test_jobs.py
```diff
@@ -1,5 +1,169 @@
 import bson
+import mock
+import pytest
 from api.jobs import job_util
+from api.web import errors
+
+
```
ryandavisflywheel: I like the explicit way these tests are written 👍

hkethi002:
Jobs can be created without satisfying required inputs via the sdk
```
project = fw.get_project('some-project-id')
gear = fw.lookup('gear/name-of-gear-requiring-an-input')
job_id = gear.run(destination=project)
```
The job will start with an empty inputs folder.

nagem:
To add to @hkethi002's comment, the original issue came from a user spawning 100s of jobs without a required license key via an SDK script.

api/jobs/job_util.py
hkethi002: nope, I'll remove that

PR #1697, opened 2019-09-18
Ad hoc fixes for multi-provider [FLYW-1535]

roycollins:
Change the ad-hoc check to be a feature that is passed into the DAO layer so it can be processed differently for each container/path that uses the model layer.  At some point we should pull that apart to make it easier to refactor smaller differing sets of functionality. 

This is a preliminary PR and I wont merge it until the business logic checks are fully verified via End to End testing. That testing is happening right now. I will update if anything has to be changed as a result of the tests.

### Review Checklist


roycollins:
Tests have passed so review as normal now.

api/dao/containerstorage.py
```diff
@@ -345,8 +346,9 @@ class AcquisitionStorage(ContainerStorage):
     def __init__(self):
         super(AcquisitionStorage,self).__init__('acquisitions', use_object_id=True, use_delete_tag=True, parent_cont_name='session', child_cont_name=None)
 
-    def create_el(self, payload, origin):
-        result = super(AcquisitionStorage, self).create_el(payload, origin)
+    def create_el(self, payload, origin, features={}):
```
ehlertjd: Generally don't use in-place dictionaries or lists for default parameters, as python will use the same *instance* for every function call, prefer to use `None` instead.

api/dao/basecontainerstorage.py
```diff
@@ -268,13 +269,29 @@ def exec_op(self, action, _id=None, payload=None, query=None, user=None,
...
+                dict: the container created
+            Raises: APIConflictException
+        """
 
-        if self.cont_name in ('sessions', 'subjects', 'analyses'):
+        if features is None:
+            # Set creation defaults here as the list grows
+            features = {'check_adhoc': False}
+
+        if features.get('check_adhoc', True):
```
ehlertjd: This behaves differently between `features=None` or `features={}`, is that intended, or should we use the same default values between the cases?

api/dao/containerstorage.py
roycollins: Good catch.  Yes those should all be features=None like they are in the other spots.  Ill get them corrected.

api/dao/basecontainerstorage.py
roycollins: Yes this is intended.  If you do not pass in features, None, then it will use default features.  If you explicitly pass in {} then no features will be set.  There could be a case where you want to skip all features and you would have to know what they are to turn them all off, or at least what the defaults are.   If you prefer a different pattern just let me know and was can change it up if it makes sense.

api/dao/containerstorage.py
roycollins: I think this comment must be from on old commit.  I didnt seen any spots that were not corrected.  Let me know if you see any spots I missed. Search replace did give any hits.

api/dao/basecontainerstorage.py
```diff
@@ -612,15 +629,19 @@ def check_adhoc(self, payload, origin):
...
         parents = self.get_parents(payload)
-        project = project_storage.get_container(parents['project'])
-        if not project['editions'].get('lab'):
+        # Check project first unless it is a project
+        if parents.get('project', None):
+            parent = ContainerStorage.factory('project').get_container(parents['project'])
+        else:
+            parent = ContainerStorage.factory('group').get_container(parents['group'])
+
+        if not parent['editions'].get('lab'):
```
ehlertjd: I'm not totally clear on the business requirements, but should this check be:
`if parent['editions'].get('lab') is None:`? 

The question being, if you explicitly have lab turned *off* for a project and *on* for a group, which setting takes effect? Should it default to the group, or honor the "off" set by the project?

roycollins: That is a good question. You need to have Lab enabled on a group in order to enable Lab on the related projects.  But you dont have to turn it on for all projects.  If that is the case then it should not flow up to the group setting if its explicitly turned off on the lab. Otherwise, you could never turn it off on the project level. But I think we should confirm this with the product owners. It is a weird state where 'Off' is the new restricted version and 'On' is the normal state.

PR #1698, opened 2019-09-19
Files metrics [FLYW-1541]

davidfarkas:
Added the following metrics:

- added new `FileCountReport` which can be used to collect file counts
- new metric `fw_core_file_count` has two labels
  - `name`: can be one of the followings: `all` and if the `virus_scan` feature is enabled: `quarantined`, `virus`, `clean`, `unchecked`
  - deleted

### Review Checklist


api/metrics/collect.py
```diff
@@ -152,10 +152,103 @@ def collect_db_metrics():
         for label, count in status_counts.items():
             values.DEVICE_STATUS_COUNT.labels(*label).set(count)
 
+        # Files count
+        cont_names = ['projects', 'subjects', 'sessions', 'acquisitions', 'analyses', 'collections']
+        files_total_cnt = 0
+        files_quarantined_cnt = 0
+        files_virus_cnt = 0
+        for container in cont_names:
+            cursor = config.db.get_collection(container).aggregate(FILES_CNT_PIPELINE)
```
ehlertjd: We may want to be aggregating these numbers as they change rather than a full aggregation query every 60 seconds.

ehlertjd: Good points @davidfarkas I missed the requirement to count *all* files in the system, and introducing deleted files also complicates things. I will bring this up with the product owners to see what our options are.

nagem: These counts were purely for our own internal monitoring, and so alerts could be set up when the difference between the rate of change for quarantined files is out of line with the rate of change of all files in the system. In addition, it would be useful to know on our end when someone has inadvertently uploaded a file marked as a "virus". 

We also noticed in another event that without search, we had no metric for the number of files in the system. 

I hope that clears the request up.

ehlertjd: Hi @davidfarkas I've updated the original issue with:

> Since tracking every file is an expensive operation, we should only be aggregating values periodically (e.g. hourly).
One possibility (in our current infrastructure) would be to use the [collect handler](https://github.com/flywheel-io/core/blob/billing-report-revived/api/reports/handler.py#L48) in the pending update to the reports API. This would allow us to use the existing cron task scheduler to periodically update file counts in a singleton document, which can then be consumed by the metrics endpoint.
> 
> In addition to the report on total, quarantined and virus file states, we should have an instantaneous counter that is incremented immediately every time a file is marked as having a virus. This counter doesn't need to reflect the total number of viruses in the system, just the viruses reported over an interval.

davidfarkas: @ehlertjd I have updated the PR, now it's based on the `billing-report-revived` branch.
One outstanding question remained. What do you mean under this? _"This counter doesn't need to reflect the total number of viruses in the system, just the viruses reported over an interval."_ specifically the _"over an interval"_ part.

api/reports/file_count_report.py
```diff
@@ -0,0 +1,78 @@
...
+    def __init__(self, params):
+        self.field_prefix = 'files'
+        self.field_suffix = 'count'
+        super(FileCountReport, self).__init__(params)
+
+    def user_can_generate(self, uid, roles):
+        """
+        User generating report must be site admin
+        """
+        has_privilege(roles, self.required_role)
```
ehlertjd: The implementation of this function should just `return False`. I don't see a need for a non-admin user to be able to generate this report, and the admin case is covered [here](https://github.com/flywheel-io/core/blob/billing-report-revived/api/reports/handler.py#L106)

api/reports/file_count_report.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+class FileCountReport(Report):
+    doc_id = 'counters'
+    required_role = Privilege.is_admin
+    can_collect = True
+
+    def __init__(self, params):
+        self.field_prefix = 'files'
+        self.field_suffix = 'count'
```
ehlertjd: Are `field_prefix` and `field_suffix` used anywhere?

api/metrics/collect.py
ehlertjd: @davidfarkas I meant that we should increase a counter metric whenever we get a callback indicating that a file has a virus.

api/reports/file_count_report.py
davidfarkas: nope, thanks, my mistake

api/metrics/collect.py
davidfarkas: @ehlertjd  It would be nice if you could rebase the `billing-report-revived` on master because this branch doesn't contain the anti-virus-check stuffs or I can do that if you give the green light.

ehlertjd: @davidfarkas I've rebased `billing-report-revived`

api/metrics/values.py
```diff
@@ -93,6 +93,10 @@
 # Number of usage report collection errors
 REPORT_COLLECTION_ERROR_COUNT = Counter('fw_report_collection_errors', 'Observed report collection errors', ['name'])
 
+# ===== Files =====
+FILE_COUNT = Gauge(prefix + 'file_count', 'Number of files in the system', ['name', 'deleted'], multiprocess_mode='max')
+
+REPORTED_VIRUSES_COUNT = Counter('fw_reported_viruses_count', 'Number of reported files having a virus', [])
```
ehlertjd: Use common prefix here (i.e. `prefix + 'reported_viruses_count'`)

PR #1699, opened 2019-09-23
add paging to gear suggestions [FLYW-1551]

roycollins:
Add paging to the gear selection endpoint. The page will take a subset up to the page size starting with parents of the level then children of the level and lastly files at the level.  Once the limit is reached it will return with what is has retrieved.  

It only makes sense at this point to add integration tests since its using all the same low level components so once its confirmed this approach looks good I will wrap the units in some integration tests.  

### Review Checklist


roycollins:
Oy, I looked at the tests for test_page and test_limit and they only check that the params are valid types.  I didn't realize those tests were so incomplete.  I will get some tighter unit tests around those components as well.  Was planning on integration tests to wrap them up but they do need more unit testing to be valuable.

roycollins:
I started mocking the function and then I found that I have to mock out 10 different dependencies so I opted for just wrapping it all in integration tests.  When we refactor the code to be more testable we can pull out the parts and do better unit tests.

api/jobs/handlers.py
```diff
@@ -276,7 +264,70 @@ def suggest(self, _id, cont_name, cid):
...
+                {'cont_type': singularize(storage.child_cont_name),
+                 '_id': c['_id'],
+                 'label': c.get('label', '')} for c in children['results']]
+
+            found_results += len(children['results'])
+            total_results += children['total']
+            if self.pagination.get('limit') and found_results >= self.pagination['limit']:
+                return response
+
+        count = self.pagination.get('limit', 100) - found_results
```
ehlertjd: We default limit to `100` here, but not above, consider getting the definitive limit & skip values at the top of the function.

api/jobs/handlers.py
```diff
@@ -276,7 +264,70 @@ def suggest(self, _id, cont_name, cid):
...
+                {'cont_type': 'analysis', '_id': a['_id'],
+                 'label': a.get('label', '')} for a in analyses['results']]
+
+        if self.pagination.get('limit') and found_results >= self.pagination['limit']:
+            return response
+
+        # Get children
+        if cont_name == 'collections':
+            # Grab subjects within the collection context
+            pagination = copy.deepcopy(self.pagination)
```
ehlertjd: Anything we can do to de-duplicate some of the boilerplate code would be good.

roycollins: I was going for the rule of three on this one.  There was not a a lot of commonality but I went ahead and refactored but was only able to save 2 lines with the new abstraction.

api/jobs/handlers.py
roycollins: The limit is only ever tested in the above cases inside an if block that validates limit exists so the default is not needed there.   I choose not to put the default at the top of the block because it was only used towards the bottom. I wanted to keep it in context as you read down to the block; close to where it's used.

PR #1700, opened 2019-09-25
Sync support [FLYW-1417]

ambrussimon:
- Added download filtering mode `"since": <epoch>`
- Added EP `/download/{ticket_id}/targets`
- Streams multipart-encoded json parts for each target within ticket
- Split out `download_targets` collection from `downloads` (previously embedded doc list)

### Review Checklist


api/data_export/mappers/download_tickets.py
```diff
@@ -6,32 +6,46 @@
...
 
         Args:
             ticket (DownloadTicket): The ticket to insert
         """
-        # For mongo, we can simply convert to dict
-        self.dbc.insert_one(ticket.to_dict())
+        ticket_dict = ticket.to_dict()
+        target_dicts = ticket_dict.pop('targets')
+        self.dbc.insert_one(ticket_dict)
+        for target_dict in target_dicts:
```
ehlertjd: This may be premature optimization, but did you consider storing up to N (e.g. 1k-10k) targets per document, rather than 1 target per document?

api/data_export/json_target_writer.py
```diff
@@ -0,0 +1,50 @@
+"""Provides functionality for writing targets as JSON from a download ticket"""
+import json
+
+from requests_toolbelt.multipart.encoder import MultipartEncoder
+
+from ..web.encoder import custom_json_serializer
+
+
+class JsonTargetWriter(object):
+    """Class that writes targets to the given tar file"""
```
ehlertjd: ~to the given tar file~ -> as a multipart response

api/config.py
```diff
@@ -271,12 +271,13 @@ def initialize_db():
     # Ref: http://api.mongodb.com/python/current/api/pymongo/collection.html#pymongo.collection.Collection.create_index
     #
     # IMPORTANT: if the TTL field is missing, documents will not expire.
-    create_or_recreate_ttl_index('authtokens',  'timestamp', 30 * 24 * 60 * 60) # 30 days
-    create_or_recreate_ttl_index('uploads',     'timestamp',       1 * 60 * 60) #  1 hour
-    create_or_recreate_ttl_index('downloads',   'timestamp',            1 * 60) #  1 minute
+    create_or_recreate_ttl_index('authtokens',       'timestamp', 30 * 24 * 60 * 60) # 30 days
+    create_or_recreate_ttl_index('uploads',          'timestamp',       1 * 60 * 60) #  1 hour
+    create_or_recreate_ttl_index('downloads',        'timestamp',            1 * 60) #  1 minute
+    create_or_recreate_ttl_index('download_targets', 'timestamp',            1 * 60) #  1 minute
```
ehlertjd: Wondering if we should increase this - is 1 minute long enough?

api/data_export/json_target_writer.py
```diff
@@ -0,0 +1,50 @@
...
+        """
+
+        def target_to_field(target):
+            if target.download_type == 'metadata_sidecar':
+                target.metadata = file_source.get_metadata(target)
+            data = json.dumps(target.to_dict(), indent=2, default=custom_json_serializer)
+            return (target.filename, (target.filename, data, 'application/json'))
+
+        fields = (target_to_field(target) for target in file_source)
+        multipart_encoder = MultipartEncoder(fields, boundary=self.boundary)
```
ehlertjd: Out of curiosity, would [Newline delimited JSON](http://jsonlines.org/) have been easier / lighter?

api/data_export/mappers/download_tickets.py
ambrussimon: I considered it indeed (as you originally suggested) - in the end I decided against it:
- the code was getting convoluted
- all resources suggested that storing/querying large collections of small docs is _the way_
- I didn't find any resources (for or against) the batch pattern with mongo
- the simple pk-fk solution translates well to any (future) mapper

This reminds me that I still need to add an index though.
Let me know if you'd still prefer batches, I'm happy to give it another go.

ehlertjd: @ambrussimon I think we'd should base any changes to this on empirical data (or recommendations from the industry). Let's leave as is, until we identify that it's truly a bottleneck and that batching is a viable alternative.

api/data_export/json_target_writer.py
ambrussimon: `.jsonl` is indeed slightly lighter (~ easier to approach for others/newcomers), while `multipart` feels more web-native. FWIW, though, `urllib3` complains when consuming proper `multipart` responses, so RFCs and web-nativity only get you so far...

api/config.py
ambrussimon: Good point - I bumped to 5 for good measure.

PR #1703, opened 2019-09-26
Resend virus scan webhooks, remove mongomock [FLYW-1542]

davidfarkas:
- Added new endpoint that can be used to resend webhooks that failed previously
- wasn't able to test the new aggregation pipeline with mongomock and since we always have difficulties with that, I removed it and unit tests use real mongodb now
- updated test requirements to the latest ones (which supports python 2)

### Review Checklist


api/triggers/handlers/virus_scan.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class VirusScanTriggerHandler(base.RequestHandler):
+    @require_privilege(Privilege.is_admin)
+    def post(self):
+        if not config.get_feature('virus_scan', False):
+            return
+
+        callback_url = config.get_config()['webhooks']['virus_scan']
+        if not callback_url:
```
ambrussimon: This piece is repeated - maybe it can be moved somewhere common?

api/triggers/handlers/virus_scan.py
```diff
@@ -0,0 +1,58 @@
...
+from ...auth import Privilege, require_privilege
+from ...dao import containerutil, liststorage
+from ...web import base, errors
+from ...webhooks import VirusScanWebhook
+from .. import mappers
+
+
+class VirusScanTriggerHandler(base.RequestHandler):
+    @require_privilege(Privilege.is_admin)
+    def post(self):
```
ehlertjd: Do we have any thoughts on how long the request might take? Should this be an SSE handler instead, to allow for indeterminate runtimes?

davidfarkas: You are right it can took some time, so I modified the handler be an SSE.

PR #1704, opened 2019-09-30
Remove requirement for AWS env vars [FLYW-1651]

roycollins:
### Review Checklist


roycollins:
Added some tests but please look them over when you get a moment. I want to be sure they are constructed in the correct fashion.

ehlertjd:
LGTM thanks

PR #1711, opened 2019-10-08
Create indices on the parent keys specifically [FLYW-1663]

hkethi002:
### Review Checklist


hkethi002:
Yeah, we should make a ticket for that, also I think the parents indices on jobs also needs to be fixed, `db.jobs.create_index('parents', **kwargs)` doesn't do what I think we want it to do.

ehlertjd:
Any reason to kick that down the road? Or can it be done quickly as part of this PR?

hkethi002:
@ehlertjd it was easy enough to add to this PR

PR #1712, opened 2019-10-09
Refactor ticket generation to reduce memory usage

ehlertjd:
### Review Checklist


ambrussimon:
@ehlertjd I tested and approve your changes, and hijacked the branch/PR to add more changes to fix memory consumption on `/download/targets`. @davidfarkas's testing on 150K files revealed that memory consumption is now in check, so with a tandem PR on CLI this should `yield` a truly scalable sync.

Note that both ticket and summary is expected to fail with huge datasets.
My proposal is to convert both to SSE.

ehlertjd:
Further changes look good to me.

Since we're using the response-handler pattern to write targets, everytime we flush the buffers we're sending data to the client (as we do in SSE), so that will reset any request timeouts in the same way that SSE does (for writing targets anyway)

Generating tickets, we'll probably want to add an SSE endpoint at some point, but I'm guessing we'll need to redesign building the tar file for downloads of that magnitude anyway.

